PR #10, opened 2015-10-17
Ensure that Wget2 does not spawn more threads than needed

### LINKED ISSUE #9: Wget2 always starts config.num_thread threads
I have outlined this issue on the Wiki page about Progress Bars. Currently, Wget2 starts config.num_threads threads irrespective of the actual number of files to download. So even if Wget2 is invoked for a single file, the main thread will spawn 5 threads (current default for num_threads) and then stop them after a NO-OP. 

This is a waste of CPU cycles. Not to mention that it is one of the issues why the progress bar is not working very well. 

Since this should be an easy fix, I'm marking it for the Initial Release milestone.

darnir:
I'm creating a pull request here since I'd like to ask for a review before actually merging this code.

The patches are simple. They do three important things:
1. Add a utils.[ch] file for defining commonly used functions.
2. Rename the --num-threads option to --max-threads. Since, we're really defining the "max" value here.
3. Use the config.num_threads variable to store the actual number of threads that the instance will use.

Here, I'm not sure if I've missed something that actually needs to use the config.max_threads option but isn't. @rockdaboot if I'm missing something, please point out.

Should fix #9

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -111,7 +111,7 @@ static int G_GNUC_WGET_NORETURN print_help(G_GNUC_WGET_UNUSED option_t opt, G_GN
 		"Download:\n"
 		"  -r  --recursive         Recursive download. (default: off)\n"
 		"  -H  --span-hosts        Span hosts that were not given on the command line. (default: off)\n"
-		"      --num-threads       Max. concurrent download threads. (default: 5) (NEW!)\n"
+		"      --max-threads       Max. concurrent download threads. (default: 5) (NEW!)\n"
 		"      --max-redirect      Max. number of redirections to follow. (default: 20)\n"
 		"  -T  --timeout           General network timeout in seconds.\n"
 		"      --dns-timeout       DNS lookup timeout in seconds.\n"
@@ -579,7 +579,8 @@ struct config config = {
 	.dns_timeout = -1,
 	.read_timeout = -1,
 	.max_redirect = 20,
-	.num_threads = 5,
+	.max_threads = 5,
+	.num_threads = 1,
 	.dns_caching = 1,
 	.tcp_fastopen = 1,
 	.user_agent = "Wget/"PACKAGE_VERSION,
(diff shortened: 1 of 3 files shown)


rockdaboot:
Thanks Darshit !
I am fine with renaming the option.

I would like to avoid function or defines like min and max.
If you declare a function min() it just works with one type of variable (in your case int).
After a while you start to define functions for long, float, double, etc... as a workaround for not having type-variable functions (well, this is just C).
Using a simple define has the problem of double evaluation, a clever one just works for GCC (typeof or __typeof__ is not ISO C).

Having two extra files for one or more kinds of min/max function is overdone in my eyes.
A simple min = a < b ? a : b; just does a very readable and obvious job.
(queue_size() is a PURE function and very likely to be inlined by -O2, if we want to make sure let's make it an explicit inline function)

This is just my opinion about min/max functions and defines.

I read your mail about indentation (I am going to answer it soon).
As long as we didn't space indented the sources, please use tab for indentation.

darnir:
> I would like to avoid function or defines like min and max.
> If you declare a function min() it just works with one type of variable (in your case int).
> After a while you start to define functions for long, float, double, etc... as a workaround for not having type-variable functions (well, this is just C).
> Using a simple define has the problem of double evaluation, a clever one just works for GCC (typeof or typeof is not ISO C).
> 
> Having two extra files for one or more kinds of min/max function is overdone in my eyes.
> A simple min = a < b ? a : b; just does a very readable and obvious job.
> (queue_size() is a PURE function and very likely to be inlined by -O2, if we want to make sure let's > make it an explicit inline function)

This is exactly the reason I wanted a review before I merge. Originally, I was going to need the min() function in multiple places. Which is when I decided to write it as a function and share it. However, eventually I simply decided to rename the num_threads option, but the min() method and utils.[ch] files stayed. 

However, I don't mind eliminating the method altogether and simply using a ternary operation there. I completely agree with your comments about type issues with the functions. Personally, I dislike the compile time defines and decided against using them in this case entirely. 

P.S.: Regarding the indentation, I'm going to have to reformat my patches then. And also reconfigure my editor to actually use tabs.

darnir:
I have updated the commits. And GitHub has automatically updated this pull request with the new ones.

ChangeLog:
1. Remove the utils.[ch] files.
2. Directly use a ternary operation to define the minimum value. 

If it looks good now, we can merge this and close another open issue.

rockdaboot:
Nice !
The commit message has a little typo (intially).
Feel free to push.

darnir:
Merged and closed #9

rockdaboot:
@darnir Sorry, I just saw a basic problem with your patch.

At the point where you calculate num_threads, you can't know how many threads will be needed
- if config.recursive is set
- if the URLs are read asynchronously from stdin (via input_thread).

darnir:
Good point! I must admit I completely forgot about the fact that Wget2 can read data asynchronously from stdin, so I hadn't considered that at all. 

Maybe, we just add an exception there for when config.recursive is set or when stdin is being used for URLs and use config.num_threads == config.max_threads in that case.

PR #11, opened 2015-10-19
Giuseppe/wget2 fixes

giuseppe:


darnir:
The reindent causes a lot of mixed indentation with tabs and spaces. The source should use either one or the other, but not a mixture. Please ensure that the re-indent is done with only tabs for example.

juaristi:
Feel free to push it once you fix the indentation issues pointed out by Darshit.

But, how easy would it be to convert the style back to something else in case we decide not to be part of GNU?

rockdaboot:
@giuseppe I merged your first commit, not sure if and how Github will react on that.

Just from the first glimpse
- Some code is now hardly readable.
- Indentations of preprocessor stuff has been removed.
- Some argument lists are garbled.
Either the code has to be re-indented with amended options or manually adjusted after indentation.

@juaristi You can't convert it back 100%. Sometimes there are reasons for special (manual) indentation or for breaking the 80 column rule (e.g. readability). An automatic tool can't know this.

giuseppe:
sorry, I thought GNU indent would not leave any tab.  I am going to rework it using also Emacs and CPPI.

I would prefer that you take a look at it before I push it, I don't feel confident in such a big change without someone else agreeing on it.

giuseppe:
another thing, can we remove all the dead code that is left commented?  It can be always restored from git, if needed. What do you think?

rockdaboot:
@giuseppe As long as we use non-GNU indentation, please don't make changes based on GNU-indentation. I can't merge them. Maybe you could write down the recipe / script so we can do the indentation easily at any point in the future.

Also, do you have a link where RMS explicitly says that official GNU software *MUST* obey the GNU coding standards (maybe they are more of a *SHOULD* ?).

Regarding commented dead code: could you please just open another issue. I would like to go through each place and see if the code should be described and kept or if it is just old remnants that can be removed.

giuseppe:
sorry, I thought we sort of agreed on making it follow the GNU Coding Standards so to make the transition from wget to wget2 easier.  The only MUST is to be a Free Software, but following the GCS helps to make it uniform with other GNU tools, either from the user POV (like having support for --help, --version) than the coding style.

juaristi:
Giuseppe is right. Here are [GNU's coding standards](https://www.gnu.org/prep/standards/html_node/index.html).

The coding style is explained in the section "[Formatting your source code](https://www.gnu.org/prep/standards/html_node/Formatting.html#Formatting)". If you look at it, it doesn't look like it's enforcing such coding style, but rather **recommending** it.

We should also take into account "[Clean use of C constructs](https://www.gnu.org/prep/standards/html_node/Syntactic-Conventions.html#Syntactic-Conventions)". It uses the words "Do" and "Don't". That's imperative so I interpret them as obligations. However, that's about the syntax, not the style.

It looks like there's no obligation to use a GNU coding style.

giuseppe:
wget is already part of GNU. If wget2 is going to replace it then I think we should use the GCS if there are no strong arguments against it that we can push.

rockdaboot:
I thought what we want is GNU without GNU indentation. Standards like --version and --help are already followed.
*if* we re-indent the code, each file has to be manually reviewed and amended.
Almost every line will be touched... that is something we don't want - only if we *MUST* obey RMS's rules to get Wget2 accepted as GNU tool we should do it.
Also, GCS says C89 Standard is ok, but pre-C89 is preferred. Haha, the code is C99 and I *really* want C99+Posix. I do not care for 0,00001% people having problems with compilation because they us 25 years old software (but want to compile modern code). These should keep on using Wget1.x.

giuseppe:
c99+posix is fine, gnulib takes care of portability and as you said, who is stuck to these old platforms should use the correct programs for them.  Currently the code is a mix of spaces and tabs, so it looks it needs to be reindented anyway. If not following the GNU coding style, what is the alternative?

rockdaboot:
>  Currently the code is a mix of spaces and tabs

It should be tabs. There are some reasonable exceptions, e.g. when wrapping lines, you might want spaces for exact column alignment.
```
<tab><tab>if (long expr...
<tab><tab>ssss&& second expression)
```
(s = space)

If you find  other occurrences of spaces for indentation, it is likely a bug.

PR #16, opened 2015-10-21
Docs

juaristi:
I documented the hashing functions (defined in hashfile.c) using GTK-Docs, which is already built into the Makefiles, and produces beautiful HTML output. Looks like some of these functions are dependent on the underlying cryptographic engine (e.g. libnettle) so I documented these on the header file (libwget.h), since there are more than one implementations. All the others are documented in the source itself (hashfile.c).

Please review my changes, specially the accuracy/correctness of the terms used. If no one complains, I'll push them next week.


giuseppe:
instead of the merge, could the branch just be rebased? Or am I missing something?

rockdaboot:
What Giuseppe says. That would allow for a straight pull without 'Merge branch 'wget2' into docs'. Not sure what happens if we merge the pull request right now.

giuseppe:
you can just do while on the branch:

git reset --hard 751540b
git pull --rebase git@github.com:rockdaboot/wget2.git wget2
git push --force git@github.com:juaristi/wget2.git docs

juaristi:
I think there's no "Merge branch 'wget2' into docs". I think what it does is to directly reply all the 4 commits in master. Anyway I'm not 100% sure either. If it doesn't work as expected, I'll do what Giuseppe says.

darnir:
Ander, what is suggest is that you do a `git merge --squash` to copy all
the commits from your docs branch to the wget2 branch. Rewrite the commit
message and then push it.

A merge --squash will allow you to cleanup your multiple commits for what
should really be only one. Then you can force update your forks.

On 21 October 2015 at 22:14, Ander Juaristi <notifications@github.com>
wrote:

> I think there's no "Merge branch 'wget2' into docs". I think what it does
> is to directly reply all the 4 commits in master. Anyway I'm not 100% sure
> either. If it doesn't work as expected, I'll do what Giuseppe says.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rockdaboot/wget2/pull/16#issuecomment-150012155>.
>



-- 
Thanking You,
Darshit Shah

rockdaboot:
So, basically to work with github pull requests smoothly: rebase instead of merge.
Creating an own branch for changes is a good idea.
I basically do the same as Giuseppe suggests (assumed you haven't done 'git merge' before):
- git checkout wget2
- git pull
- git checkout docs
- git rebase wget2

Of course I could easily merge your commits manually (using git remote add and git merge), but things like this are always a good exercise.

rockdaboot:
I cherry-picked your commits, that means history is broken and I have to close this pull request by hand. Also you should either re-create your branch 'docs' or rebase it on branch 'wget2'.

PR #18, opened 2015-10-21
OCSP and HSTS files should be stored to ~/.wget_*

### LINKED ISSUE #14: Wget2 creates .wget_ocsp and .wget_ocsp_hosts files in $PWD
As I understand these files are to allow Wget to realize during a future run which certificates it has already checked. However, it always creates these files in the $PWD while they should be in $HOME

darnir:
Currently the ocsp and hsts files are being created in the $PWD. Instead initialize them to point to a constant location in $HOME.
Will close #14

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -48,7 +48,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <ctype.h>
-//#include <pwd.h>
+#include <pwd.h>
 #include <errno.h>
 #ifdef HAVE_GLOB_H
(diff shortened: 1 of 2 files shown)


rockdaboot:
Nice.
You could remove the explicit '... = NULL' from config initialization and then merge the commits.

darnir:
Merged.
I kept the explicit NULL assignments because I thought that we're initializing all values there. Removed the lines before merging.

Closes #14

PR #19, opened 2015-10-22
giuseppe/wget2 fixes

giuseppe:


rockdaboot:
Just some questions/comments:
I often make lists of functions with the same return type, e.g.
```
void
  func_a(),
  func_b(),
  func_c();
```

Why do you think this has to be fixed ? Is it just a matter of taste or is there some practical reason not obvious to me ?

In libwget/http.c the fix should be
```
  iri = wget_iri_parse (host, encoding);
  if (iri)
    wget_vector_add_noalloc(proxies, iri);
  else
    print warning
```
instead of
```
  iri = wget_iri_parse (host, encoding);
  if (!iri) {
    wget_vector_free (&proxies);
    return NULL;
  }
  wget_vector_add_noalloc(proxies, iri);
```
The reason is, that we parse a comma-separated list of proxies. If one has a bad syntax, we should just print a warning and go on.

giuseppe:
probably the main reason is that I got used to see the function signature complete and I found slightly difficult to see immediately the return type.  But if this is not convincing you, feel free to drop the patch :)

About the warning, I think we should bail out if the user is providing the wrong input, wget is going to be used widely in scripts and non interactive scenarios, an error is easier to catch than a warning, especially if it happens immediately after the program starts.

rockdaboot:
Regarding function declarations, I am fine with your patch. GTK-DOC might have problems with comma-separated lists. But BTW, how can I drop one of your commits and merge the others (without breaking the history) ?

Proxies: An error message is fine. But your code lets wget2 continue without proxies at all - that might be very dangerous for the user.

giuseppe:
what I do is to fetch locally the commits, I do that also for reviews, then git rebase -i and unselect the patch you don't want to include.  Then the PR must be closed manually

giuseppe:
OK to add another patch that propagate the error to the upper layers?

rockdaboot:
Yes. I guess it is good to propagate the error to upper layers.

juaristi:
I also made similar changes in #16 . The reason is just what you said. GTK-Doc does not like comma-separated declarations. I don't remember right now what was the exact issue, but I do remember something went wrong with GTK-Doc that got fixed once I re-declared the functions the "normal" way.

giuseppe:
pushed two more patches

rockdaboot:
Thanks, all commits rebased and pushed.

PR #33, opened 2015-10-30
Avoid potential race condition

### LINKED ISSUE #23: Fix potential race condition in HSTS
Same as [f5a63e31006ed02f5de4bb8d947bc7b755bb81fe](http://git.savannah.gnu.org/cgit/wget.git/commit/?id=f5a63e31006ed02f5de4bb8d947bc7b755bb81fe) in wget.

I have seen no file locking in the HSTS code. While it's true it creates some (one?) mutexes, they seem to be for internal threads synchronization only (i.e. to make the HSTS API thread-safe).

Although I haven't had time to prove this yet, it looks like other wget2 processes might clobber the HSTS file.

juaristi:
This fixes #23.


juaristi:
Latest changes:
 - If the HSTS table is empty, a new debug message is printed saying: "No HSTS entries to save. Table is empty.\n".
 - If ftruncate fails, we try to close and reopen the file only for writing ("w").

juaristi:
Why use a temp file? Why not write directly to the HSTS file?

The question now is, how should the lock file be called? What if another file with the same name already exists? We either lock that file (even if it's not "ours") or we create a unique file (by appending ".1", ".2", etc. as wget does) and use it as the lock. But neither of them seem to be a good idea for me. In the first approach, you'd lock a file that other process might be using. In the second approach, other processes will create unique lock files as well (".2", ".3", ".4", etc.), so there would be no locking at all.

Maybe isn't just better to forget about it and fail if `ftruncate()` fails (as we do with `fclose()`)?

rockdaboot:
> Why use a temp file? Why not write directly to the HSTS file?

It about not possibly destroying existent data.

*Explanation*

Assume that ftruncate() succeeds, but the following write(s) fail. All the precious HSTS data is lost in this case. I can think of a few situations where this might happen (pressing CTRL-C, Disk full, sudden NFS networking down, ...). To avoid this possibility, we first write into a temp file (unique filename, *must* be one the same partition - best place is the same directory as the HSTS file). Only if this succeeded, we rename() it to .wget_hsts. If the rename fails - no problem, we still have the old data (and optionally may keep the new file and print a hint to the user).

juaristi:
Understood ;D
Now, what about the lock file? What do we do if a file with the same name already exists?

rockdaboot:
Tmpfile() or mkstemp() should be the solution. I guess the latter  is better  suited To create a file in the homedir. I am just in the Train ... Hard To type Here, german  auto correction 😕

rockdaboot:
I made up a function, I should finish it tomorrow so that you can use it (too many customers called today).
We will also use it for OCSP and Cookie file updating.

juaristi:
I updated this. Please have a look when you can.

rockdaboot:
Good work !
Merging pull requests via Github f*cks up our history with non-GNU-like commit messages.
So I decided to manually merge your changes and changed github repo history - so please fetch/pull with -f, else you'll see errors. Sorry for the extra work :-(

I also made OCSP and HSTS code a bit more similar (just some amendments to your code).

PR #34, opened 2015-11-03
Fixed HSTS policy application

juaristi:
I don't know if I'm missing something, but from my tests, HSTS policies didn't seem to be applied.

Before the patch (latest git):
```
$ src/wget2 --delete-after http://www.airbnb.es
[0] Downloading 'http://www.airbnb.es' ...
HTTP response 301 Moved Permanently
[0] Downloading 'https://www.airbnb.es/' ...
WARNING: OCSP is not available in this version of GnuTLS.
HTTP response 200 OK
```

After the patch:
```
$ src/wget2 --delete-after http://www.airbnb.es
[0] Downloading 'http://www.airbnb.es' ...
HSTS in effect for www.airbnb.es:80
WARNING: OCSP is not available in this version of GnuTLS.
HTTP response 200 OK
```
I've effectively checked how after my patch, the website was loaded directly through HTTPS.
All the tests were performed with an already existing `~/.wget_hsts` for `www.airbnb.es`.

### DIFF
--- a/include/libwget.h
+++ b/include/libwget.h
@@ -852,6 +852,8 @@ char *
 	wget_iri_get_query_as_filename(const wget_iri_t *iri, wget_buffer_t *buf, const char *encoding) G_GNUC_WGET_NONNULL((1,2));
 char *
 	wget_iri_get_filename(const wget_iri_t *iri, wget_buffer_t *buf, const char *encoding) G_GNUC_WGET_NONNULL((1,2));
+const char *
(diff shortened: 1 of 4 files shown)


rockdaboot:
Cool. Good idea to add set_scheme function.
Optionally you may remove the extra spaces in
+	* const iri_ports[]   = { "80",   "443" };	// default port numbers for the above schemes

else go and merge it !

juaristi:
Merged!

PR #41, opened 2015-11-30
Inline docs & refactorings

juaristi:
One final thing that worries me is how to handle the un/signedness of read/fread return values.

### DIFF
--- a/docs/libwget/libwget-sections.txt
+++ b/docs/libwget/libwget-sections.txt
@@ -43,6 +43,7 @@ wget_ready_2_read
 wget_ready_2_write
 wget_ready_2_transfer
 wget_read_file
+wget_update_file
 </SECTION>
 
 <SECTION>
--- a/libwget/buffer.c
+++ b/libwget/buffer.c
@@ -18,7 +18,7 @@
  * along with libwget.  If not, see <http://www.gnu.org/licenses/>.
  *
  *
- * Memory buffer datastructure routines
+ * Memory buffer data structure routines
  *
  * Changelog
  * 22.08.2012  Tim Ruehsen  created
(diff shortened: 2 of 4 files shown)


rockdaboot:
Cherry-picked 23fcb96 (the other commits are already in the 'wget2' branch).
Please remove your branch and start a new one from 'wget2' before you make any new changes.
Your branch is out-of-sync with 'wget2'.

juaristi:
What happens with the signedness difference in read() and fread() return values?

rockdaboot:
Well, we support just max 2GB long files :-) else reader() should return size_t.
Due to while(reader()>0) it seems ok to me. Error or end-of-file is detected.
Maybe a bit cleaner would be to check ferror() in __read() and return -1 on error.

juaristi:
Yes, I guess it's cleaner. After the while loop we have
```
if (nbytes == -1 && errno != EAGAIN)
```
and since `fread()` returns 0 even in case of failure, it wouldn't enter the if, therefore not reporting the failure to the user. We check for the length later, but the user wouldn't know something bad happened, which is IMO something important.

Please check out this:
```
diff --git a/libwget/io.c b/libwget/io.c
index 679f800..41b92f4 100644
--- a/libwget/io.c
+++ b/libwget/io.c
@@ -54,7 +54,11 @@
 
 static ssize_t __read(const void *fp, char *dst, size_t len)
 {
-	return fread(dst, 1, len, (FILE *)fp);
+	FILE *f = (FILE *)fp;
+	size_t ret = fread(dst, 1, len, f);
+	if (ferror(f))
+		return -1;
+	return ret;
 }
 
 static ssize_t __readfd(const void *f, char *dst, size_t len)
-- 
```
I don't know why, but when I preview this message GitHub converts tabs to spaces, please take it into account when merging.

rockdaboot:
Thanks, slightly amended and pushed

PR #42, opened 2016-01-05
Fix combination of -c and -O

darnir:
* src/wget.c (http_get): Stat the output document if -O is provided
* src/options.c (init): Do not truncate the output document if user has
asked to continue downloads


rockdaboot:
Do you mind to add put test into tests/test-wget-1.c ? There are already some tests using -O, so it should be straight forward.

darnir:
Oh yes, I should do that. I simply tried fixing the issue in the free time
I had.

Will add the tests. Also trying to get Travis working for Wget2 on osx.

Thanking You,
Darshit Shah

darnir:
I'll open a new request later with tests

Issue #45, opened 2016-01-07
Tests do not execute correctly


darnir:
Multiple issues when trying to run the test suite.

1. `make check` does not execute the tests correctly. It seems to be skipping some tests without a reason
2. Test results change based on the `$PWD` from which the test was executed.

Both the tests seem to be related to the same issue. The path to the PSL data file is relative, which causes these problems. 

There's also another issue that where `./tests/test` prints `(null)` as the result of two tests. This should not happen, and if it does, should be caught by the test suite at some point.

As of now a few PSL tests seem to be failing, but `make check` reports no errors, because it does not run the tests at all.

rockdaboot:
> There's also another issue that where ./tests/test prints (null) as the result of two tests. This should
> not happen, and if it does, should be caught by the test suite at some point.

I am just working on the test suite (hashing failure with libnettle weren't caught).

> make check does not execute the tests correctly. It seems to be skipping some tests without a reason

Some of the FTP tests ? Not sure what tests you are talking about - how to reproduce ?

darnir:
No, these are HTTP tests. In `./tests/test.c` the PSL data is not correctly initialized. Hence, those tests fail. But the file isn't able to catch the failures for some reason. For example, the contents of `./tests/test.log` is:

    0 XML, 0 HTML and 0 CSS files parsed
    (null)
    (null)
    (null)
    (null)
    ID=65=abcd
    ID=65=abcd
    ID=65=abcd
    (null)
    ID=65=abcd
    Summary: All 194984 tests passed
  PASS test (exit status: 0)

Whereas, the output generated when I execute `./tests/test` is:

    0 XML, 0 HTML and 0 CSS files parsed
    Failed [0]: PSL check(ID=65=abcd; expires=Tuesday, 07-May-2013 07:48:53 GMT; path=/;domain=.example.com; HttpOnly) -> -1 (expected 0)
    Failed [1]: PSL check(ID=65=abcd; expires=Tue May 07 07:48:53 2013; path=/; domain=.example.com) -> -1 (expected 0)
    Failed [2]: PSL check(ID=65=abcd; expires=Tue, 07-May-2013 07:48:53 GMT; domain=.example.com) -> -1 (expected 0)
    Failed [3]: PSL check(ID=65=abcd; expires=Tue, 07-May-2013 07:48:53 GMT; path=/) -> -1 (expected 0)
    Failed [4]: PSL check(ID=65=abcd) -> -1 (expected 0)
[...]

PR #47, opened 2016-01-08
Fix tests

darnir:
These are the changes I'd like to make wrt #45 

Is keeping the printf necessary at all? Or can we simply eliminate it entirely?


rockdaboot:
I removed the unneeded printing and pushed your changes.

PR #55, opened 2016-01-19
Fix -O/dev/null

giuseppe:


rockdaboot:
Thanks, merged both via command line.

PR #59, opened 2016-02-02
Fix --disable-threads

giuseppe:


rockdaboot:
Good work ! (And even better... downloads work without threading :-))

We should gnulib module stdbool later and could use G_GNUC_WGET_UNUSED for the unused function params to avoid warnings.

PR #72, opened 2016-03-24
http: fix multi part files

giuseppe:
* src/wget.c (download_part): Read temp part for the temporary files.
(_get_header): Write part file to a temporary file.
(http_get): Keep a temporary fd for the file parts.
(struct _body_callback_context): Add 'head' and 'part'.
* src/job.h (struct PART): Add 'tmp_fd'.
* src/job.c (job_create_parts): Initialize tmp_fd.
* autogen.sh (libwget_gnulib_modules): add 'safe-read'

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/src/job.h
+++ b/src/job.h
@@ -43,6 +43,7 @@ typedef struct {
 	char
 		inuse,
 		done;
+
 } PART;
 
 struct JOB {
--- a/src/wget.c
+++ b/src/wget.c
@@ -47,7 +47,7 @@
 #include <locale.h>
 #include "timespec.h" // gnulib gettime()
 
-#include "safe-write.h" // gnulib gettime()
+#include "safe-write.h"
 
 #include <libwget.h>
 
@@ -2161,21 +2161,8 @@ int download_part(DOWNLOADER *downloader)
 					print_status(downloader, "part %d download error '%zd bytes of %lld expected'\n",
 						part->id, resp->body->length, (long long)part->length);
 				} else {
-					int fd;
-
 					print_status(downloader, "part %d downloaded\n", part->id);
-					if ((fd = open(metalink->name, O_WRONLY | O_CREAT, 0644)) != -1) {
-						ssize_t nbytes;
-						if ((nbytes = pwrite(fd, resp->body->data, resp->body->length, part->position)) == (ssize_t)resp->body->length)
-							part->done = 1; // set this when downloaded ok
-						else
-							error_printf(_("Failed to pwrite %zd bytes at pos %lld (%zd)\n"), resp->body->length, (long long)part->position, nbytes);
-
-						close(fd);
-					} else {
-						error_printf(_("Failed to write open %s\n"), metalink->name);
-						set_exit_status(3);
-					}
+					part->done = 1; // set this when downloaded ok
 				}
 
 				wget_http_free_response(&resp);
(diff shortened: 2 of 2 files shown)


rockdaboot:
You introduce a lot of overhead (copying via read / write).
I really liked the elegant open/pwrite/close code.

Do you see a chance to improve the logic ? 
At least a buffersize of 4096 is pretty small - having chunks of several MB in mind when downloading large files over a fast network (1Gbit/s at home and 10Gbit/s for server farms is does already exist). Disk I/O and/or CPU (context switches) should not be a bottle neck here.

Could 'sendfile' be a helper at least for Linux systems ? Moving the read/write loop into the kernel...

giuseppe:
in theory most of the file content will be kept in memory by the operating system so we don't have to worry about caching for small files.  At least now it will be possible to download a huge file without keeping it all in memory.  I have no numbers but we can tune the buffersize.

Last time I checked sendfile was not supporting the copy from a file to file.

rockdaboot:
> In  Linux  kernels before 2.6.33, out_fd must refer to a socket.  Since Linux 2.6.33 it can be any file.  > If it is a regular file, then sendfile() changes the file offset appropriately.

Also:
> The copy_file_range() system call first appeared in Linux 4.5.

Lol, I don't even have 4.5 running. But the man pages are there...

BTW, --chunk-size already supports writing the chunks onto disk. I'll have a look into the code if we can easily reuse it for our purpose. (Or did you already examine it ?). And I'll make a few tests with your patch and add it to https://github.com/rockdaboot/wget2/wiki/Speed-and-CPU-measurements

rockdaboot:
Your patch increased sys time by a factor of more than 2. I'll do more testing at the weekend. Should be possible to tune it.

giuseppe:
what tests did you run?  I have used the http://ftp.debian.org/debian/pool/main/lib3/lib3ds/lib3ds-dev_1.3.0-6_ia64.deb you provided me and I get similar results (but the file is too small to give any sense to numbers)

rockdaboot:
Look here: https://github.com/rockdaboot/wget2/wiki/Speed-and-CPU-measurements
> Local network, 100Mbps, Single file, 465453147 bytes of gzipped data

Local network, not localhost. But that shouldn't matter. The other machine had a stock Apache on a stock Debian stable. But that shouldn't matter as well.

The commands I used are in the table.

You can easily generate a file with dd if /dev/random ... , zipping is not needed.

giuseppe:
thanks for the information.

What versions are exactly "Before Giuseppe's patch" and "After Giuseppe's patch"?

I guess it is before and after a181868655f1b201f2e8ae078503c0a20e59b33b as you measured the single threaded case?

In this case, most of the difference comes in my testing from the additional `fsync` call which is not done in a181868655f1b201f2e8ae078503c0a20e59b33b.

rockdaboot:
> I  guess it is before and after a181868 as you measured the single threaded case?

Yes, before that commit. And after the latest patch of this PR.

> In this case, most of the difference comes in my testing from the additional fsync call

The fsync is just called once, after the download finished, right ? So I guess it's more the additional writes().

I am not in the office the next 10 days or so. I try to set up a test environment at home (maybe via loopback or VM). My current plan is to reduce the enormous amounts of calls to read() (one read for each 1440 bytes - you can see with wget2 -d). I'll try recvmmsg() in the next day. Maybe you could strace or valgrind --tool=callgrind to see if it is fsync or write or any thing else and try to optimize that !? One approach could be to do writes less often. E.g. max every 0.1s or at 1MB barriers whichever comes first. On the first release it would be a joy if we could say: much less CPU utilisation than comparable tools (wget, curl, ari2, ...). At least for plain text connections.

giuseppe:
I changed the code now so that there is not a small buffer used for reading and writing but it is written directly as new data is coming.  I tried without the fsync call and I get similar results as before the series.  I think this one patch can be merged as it is now as it fixes the broken multipart case in a181868.

rockdaboot:
Thanks, cherry picked.

rockdaboot:
There seems to be another bug with your patch and metalink.
Downloading a metalink file leaves the intermediate XML file on the disk... fixing by unlinking seems not elegant - we don't need the XML file on disk at no time.
Example file: http://go-oo.mirrorbrain.org/stable/linux-x86/3.2.1/ooobasis3.2-af-calc-3.2.1-9505.i586.rpm

giuseppe:
thanks for the feedback.  Fixed in: https://github.com/rockdaboot/wget2/pull/73

PR #73, opened 2016-03-26
src/wget.c (_get_header): Do not store metalink files

giuseppe:
Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


rockdaboot:
Thanks !

PR #74, opened 2016-03-28
wget: new option --follow-metalink

giuseppe:
* src/options.h (struct config): Define follow_metalink.
* src/options.c (print_help): Add description for follow-metalink.
(struct options): Add follow-metalink.
(struct config): Initialize follow_metalink.
* src/wget.c (downloader_thread): Skip following metalinks when follow_metalink is off.
(_get_header): Store the metalink file when follow_metalink is false.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -191,6 +191,7 @@ static int G_GNUC_WGET_NORETURN print_help(G_GNUC_WGET_UNUSED option_t opt, G_GN
 		"      --content-disposition  Take filename from Content-Disposition. (default: off)\n"
 		"      --default-page      Default file name if name isn't known. (default: index.html)\n"
 		"      --netrc-file        Set file for login/password to use instead of ~/.netrc. (default: ~/.netrc)\n"
+		"      --follow-metalink   Follow a metalink file instead of storing it (default: on)\n"
 		"\n");
 	puts(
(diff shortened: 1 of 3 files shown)


rockdaboot:
Very good ! I'll merge it when back on the desktop. Wdyt about enabling it with --mirror ?

giuseppe:
good idea.  I amended it in the new version

rockdaboot:
WDYT about renaming --follow-metalink into --metalink ?
Also, we need a --force-metalink <file>. Do you plan such a thing or should I add it ?

giuseppe:
I am fine with renaming it.  What would --force-metalink do?  How will it be different than --input-metalink?

rockdaboot:
`--input-metalink` is IMO badly named (it is currently wget1.x only).
The way wget1/2 specifies file types is via --force-<type> --input-file <file>.
To be consistent, we should have --force-metalink and not --input-metalink <file>.

On the other hand side, we could obsolete the combination of --force-<type> and --input-file and replace it by --input-<type>. Less to type, easier to read and memorize. And makes it possible to specify multiple files with different types on one command line.

rockdaboot:
Meanwhile I added --force-metalink and made some metalink code cleanups.

PR #77, opened 2016-06-16
Add compiler sanitizer options

darnir:
Add some compiler -fsanitize options:

1. Undefined Behaviour Sanitizer (ubsan): Detect issues in the code that are undefined behaviour according to the C specification
2. Address Sanitizer (ASan): Detect addressing issues and memory leaks at runtime. Slightly similar to Valgrind, but using compiler added instrumentation

There's a couple of caveats to this set of patches:

1. I've added the `-fsanitize` flags directly to the GCC Warnings in `configure.ac`. This means that any downstream maintainer that decides to enable warning flags will unknowingly enable these sanitization options as well. These options add instrumentation instructions to the compiled binary causing significant slowdown and will also cause crashes in many cases. As developers we want this, but end-users don't. We need to set up a new configure flag to enable such options.

2. These don't seem to be working on Travis. The version of GCC is too old on travis to support these sanitization options. And with Clang, for some reason, Wget2 simply doesn't fail. On my system, the Address Sanitizer shows a couple of issues with the metalink test, but everything passes fine on Travis. I'll look deeper into it later, especially given how Travis currently has a few problems with newer Clang versions

Since this is turning to multiple small commits, I'll re-create this into a larger single patch after comments and suggestions

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ./autogen.sh || exit 1
-./configure || exit 1
+./configure --enable-gcc-warnings || exit 1
 make -j3 || exit 1
 if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
   make install || exit 1
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ./autogen.sh || exit 1
-./configure --enable-gcc-warnings || exit 1
+./configure --enable-gcc-warnings --enable-fsanitize || exit 1
 make -j3 || exit 1
 if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
   make install || exit 1
--- a/configure.ac
+++ b/configure.ac
@@ -176,6 +176,8 @@ if test "$gl_gcc_warnings" = yes; then
     gl_WARN_ADD([$w])
   done
   gl_WARN_ADD([-fdiagnostics-show-option])
+  gl_WARN_ADD([-fsanitize=undefined])
+  gl_WARN_ADD([-fno-sanitize-recover=undefined])
   gl_WARN_ADD([-Wno-missing-field-initializers]) # Rely on missing field = 0.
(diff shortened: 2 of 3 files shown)


rockdaboot:
Nice, looks good. I'll test it tonight.
For the too old GCC, there is a GCC version check in configure.ac that you can use. There is already an example.

darnir:
I saw that. That's not an issue since using `gl_WARN_ADD` adds a configure rule to test the compiler capability before using it. The issue is that we don't see the problems on travis tests.

rockdaboot:
Go ahead.
The issue with test-metalink are allocations that are freed before exit (on purpose).
Though I have to look at realloc() for downsizing allocated memory... seems to be odd why something is lost here.
Using valgrind directly on that test prints out the same errors, so I guess it is correctly stated.

darnir:
I'll send a new pull request with only two commits. One with all the sanitizers in a new configure option and one that fixes the ubsan detected errors.

However, one question, do you know how to compile two different executables during make check? Because the address sanitizer and memory sanitizer cannot be passed together. There's two ideas for that:

1. Create two configure options that are mutually exclusive and the travis script compiles and checks both serially.
2. Make check compiles both and executes them

The first one is easy, bit relies a lot on travis identifying issues. However, I'm not sure how to implement the second method.

rockdaboot:
Regarding 2. : see src/Makefile.am where we already compile wget2 + wget2_noinstall (that is used for the test suite).

darnir:
I saw, will take a deeper look later. This requires building all the applications in tests/ with a different name and different CFLAGS and only when the right configure flags are passed. This seems a little too complex and difficult to maintain.

rockdaboot:
If I understood correctly, you have to completely recompile with two different sets of compiler flags anyway. That includes libwget as well. So, your choice 1 seems to fit perfectly. Sorry, if my last comment led you into the wrong direction...

rockdaboot:
@darnir What is the status with this ?

PR #84, opened 2016-07-22
Fix issue with --continue and -O are passed together

darnir:


darnir:
Closing this in favour of #85

PR #87, opened 2016-07-22
Print filename in progress bar

darnir:
This is a very hacky commit. Please give your reviews on improvements and then I'll make a new pull request with the full changelog and polished code.

### DIFF
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -56,6 +56,7 @@ glob
 iconv
 inline
 inttypes
+ioctl
 lib-symbol-visibility
 listen
 malloc-posix
--- a/include/libwget.h.in
+++ b/include/libwget.h.in
@@ -312,6 +312,8 @@ int
 	wget_match_tail(const char *s, const char *tail) G_GNUC_WGET_PURE LIBWGET_EXPORT;
 int
 	wget_match_tail_nocase(const char *s, const char *tail) G_GNUC_WGET_PURE LIBWGET_EXPORT;
+char *
+	wget_human_readable (size_t n, const int acc, const int decimals) G_GNUC_WGET_PURE LIBWGET_EXPORT;
 ssize_t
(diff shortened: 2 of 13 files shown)


darnir:
This pull request is not ready for merging yet. I wanted some review on the code style for implementing the feature. I'm currently also working on some more progress bar changes. Will keep pushing to this branch so everyone can take a look.

If there is any concern, please flag it immediately so I can rectify.

rockdaboot:
I just can go step-by-step through the commits... let's start with 859c21b.

I accidentially left 'printf' in libwget/bar.c. IMO, it should be (wget_)info_printf.
That gets us the opportunity to add a new format specifier e.g. %Wh. (W = special Wget format specifier, h = human readable with size_t argument expected). Feel free to propose something else.
WDYT ?

rockdaboot:
ioctl(): Add module ioctl to bootstrap.conf. Not sure how much compatibility gnulib provides - but it is a chance. Look at ./boostrap output regarding which file names to include.

rockdaboot:
It looks good - didn't test it though.
I hope that a multi column bar turns out to be the right design decision - not 100% sure yet.
But on the other side... we want to go new ways and push everything to the limit :-)

rockdaboot:
Ähm, could you extend the commit messages with the files changed ? I guess it is still useful when generating the ChangeLog from the git commit messages.

darnir:
Nope. I think the `printf()` call there is correct. The logger disabled printing "info" messages when the progress bar is enabled. Hence, we either print the progress bar directly using `printf()` or create a new logging level that allows us to print messages unconditionally  (like DEBUG, but without the fancy colors) to the screen.

I'll add the ioctl module soon as well. About the commit messages, these commits are only temporary and proof of concept. I'll clean them up with proper commit messages once the design and implementation have been figured out.

darnir:
Is it okay to close this in favour of #90 ? Or would you like to merge filename support first and look at the asynchronous bar later?

rockdaboot:
If filename support is also in #90, just close it.

PR #88, opened 2016-07-23
Add compiler sanitizer options

darnir:
This set of commits is an updated version of #77. It adds three sanitization options: 

  1. Undefined Behaviour
  2. Address Sanitizer
  3. Memory Sanitizer

These still don't work on Travis, but work perfectly fine locally. As mentioned earlier, the version of GCC on Travis is too old to support these options. While with Clang, for some reason, configure simply does not enable these options. I'll take a deeper look into why this happens later. However, these commits should still be reviewed and merged on their own.

Closing #77 in favour of this branch

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ./bootstrap || exit 1
-./configure || exit 1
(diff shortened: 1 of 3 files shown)


rockdaboot:
Rebased + merged

PR #90, opened 2016-07-28
Create an asynchronous progress bar

darnir:
Give the progress bar its own thread and let it update the entire
display asynchronously at a specified time interval. The existing
implementation refreshes the progress bar for each network packet
downloaded. Over multiple downloader threads and a high speed network
connection this can lead to far too many redrawings of the screen. Also,
each of the downloader threads will block while trying to acquire the
thread mutex because another thread just retrieved a packet. While I
haven't profiled it, it seems like there would be extremely high lock
contention in the existing implementation. Instead, with a separate
thread, we can update all the progress bar slots simultaneously at
regular intervals.

    * bootstrap.conf: Include Gnulib module "ioctl"
    * include/libwget.h.in: Define always_inline, flatten and deprecated
	compiler attributes for both GCC and Clang
    * include/libwget.h.in: Export new functions wget_human_readable,
	wget_bar_register and wget_bar_deregister. Also remove unused
	function wget_bar_update
    * include/libwget.h.in (wget_http_response_t): Add new element
	"cur_downloaded" to struct. This element keeps a track of the raw
	number of bytes downloaded in the response
    (_wget_bar_ctx): Define new struct for storing the progress bar
	context.
    * libwget/bar.c: Fix display of downlaoded ratio. Ensure it does not
	exceed 100%.
	Add support for printing the filename and downloaded bytes to the
	progress bar.
	Create a new progress bar thread on initialization
    * libwget/http.c (wget_http_response_cb): Store the raw number of
	bytes downloaded from the network in the response data
    * libwget/utils.c (wget_human_readable): New function to convert an
	int value to a human readable string
    * src/bar.c: Use the entire screen width instead of just 70 columns
	for the progress bar.
	Provide methods to register and deregister a progress bar
	context
    * src/bar.h: Same
    * src/job.h: Move definition of struct DOWNLOADER from main.c to
        here
    * src/progress.h: Move definition of _body_callback_context from
        main.c to here
    * src/utils.c: Add new method to determine the width of the screen
    * src/utils.h: Same
    * src/wget.c: Use the new progress bar contexts. Update, register
    and deregister them
    * src/Makefile.am: Add new files, progress.h, utils.c, utils.h

### DIFF
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -91,6 +91,7 @@ sendto
 servent
 setlocale
 setsockopt
+signal-h
 socket
 socklen
 stdarg
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -56,6 +56,7 @@ glob
 iconv
 inline
 inttypes
+ioctl
 lib-symbol-visibility
 listen
 maintainer-makefile
--- a/include/libwget.h.in
+++ b/include/libwget.h.in
@@ -741,6 +741,8 @@ void
 int
 	wget_thread_kill(wget_thread_t thread, int sig) LIBWGET_EXPORT;
 int
+	wget_thread_cancel(wget_thread_t thread) LIBWGET_EXPORT;
+int
 	wget_thread_join(wget_thread_t thread) LIBWGET_EXPORT;
 int
 	wget_thread_cond_init(wget_thread_cond_t *cond) LIBWGET_EXPORT;
--- a/include/libwget.h.in
+++ b/include/libwget.h.in
@@ -1846,6 +1846,8 @@ void
 	wget_bar_register(wget_bar_t *bar, wget_bar_ctx *ctx) G_GNUC_WGET_NONNULL_ALL LIBWGET_EXPORT;
 void
 	wget_bar_deregister(wget_bar_t *bar, wget_bar_ctx *ctx) G_GNUC_WGET_NONNULL_ALL LIBWGET_EXPORT;
+void
+	wget_bar_update(const wget_bar_t *bar, int slotpos) G_GNUC_WGET_NONNULL_ALL LIBWGET_EXPORT;
 
 
 WGET_END_DECLS
--- a/include/libwget.h.in
+++ b/include/libwget.h.in
@@ -118,9 +118,15 @@
 
 #if GCC_VERSION_AT_LEAST(3,1)
 #	define G_GNUC_WGET_ALWAYS_INLINE __attribute__ ((always_inline))
+#   define G_GNUC_WGET_FLATTEN __attribute__ ((flatten))
+#   define G_GNUC_WGET_DEPRECATED __attribute__ ((deprecated))
+#elif defined(__clang__)
+#   define G_GNUC_WGET_ALWAYS_INLINE __attribute__ ((always_inline))
+#   define G_GNUC_WGET_FLATTEN __attribute__ ((flatten))
 #	define G_GNUC_WGET_DEPRECATED __attribute__ ((deprecated))
 #else
 #	define G_GNUC_WGET_ALWAYS_INLINE
+#	define G_GNUC_WGET_FLATTEN
 #	define G_GNUC_WGET_DEPRECATED
 #endif
 
@@ -312,6 +318,8 @@ int
 	wget_match_tail(const char *s, const char *tail) G_GNUC_WGET_PURE LIBWGET_EXPORT;
 int
 	wget_match_tail_nocase(const char *s, const char *tail) G_GNUC_WGET_PURE LIBWGET_EXPORT;
+char *
+	wget_human_readable(size_t n, const int acc, const int decimals) G_GNUC_WGET_CONST LIBWGET_EXPORT;
 ssize_t
 	wget_fdgetline(char **buf, size_t *bufsize, int fd) G_GNUC_WGET_NONNULL_ALL LIBWGET_EXPORT;
 ssize_t
@@ -1499,6 +1507,8 @@ struct wget_http_response_t {
 		hsts_include_subdomains;
 	unsigned char
 		hsts : 1; // if hsts_maxage and hsts_include_subdomains are valid
+	size_t
+		cur_downloaded;
 };
 
 typedef struct {
@@ -1803,7 +1813,22 @@ void
  * Progress bar routines
  */
 
+struct _wget_bar_ctx {
+	off_t
+		slotpos;
+	size_t
+		expected_size,
+		raw_downloaded;
+	const char *
+		filename;
+	int
+		final;
+	wget_thread_mutex_t
+		mutex;
+};
+
 typedef struct _wget_bar_st wget_bar_t;
+typedef struct _wget_bar_ctx wget_bar_ctx;
 
 wget_bar_t *
 	wget_bar_init(wget_bar_t *bar, int nslots, int max_width) LIBWGET_EXPORT;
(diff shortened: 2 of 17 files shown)


darnir:
Any reviews or comments on this?

rockdaboot:
Sorry, remember I saw 'UnderConstruction' and assumed you are still working on it... I'll have a look at it before the weekend.

rockdaboot:
Redirecting the output should switch off the bar. Except someone says --progress=bar,force. WDYT ?

IMO, the bar code in libwget should be a basic API / progress helper functions. You put the thread code in there... Can't that stay in src/bar.c. It seems very specific to wget2.c. Was that intentional ?

rockdaboot:
But yeah, I like the principle and the overall code looks very good !

darnir:
I guess it would make more sense to split out threading code and put it in `src/wget.c`. That would let us keep `libwget/bar.c` as only helper functions.  I'll work on that front next.

The reason I added the "underContstruction" tag was to indicate that it isn't ready for merging yet. This request is only for code review.

darnir:
I moved some code around so that all the threading is now done in wget2 instead of libwget. I haven't tested it very thoroughly, but initial tests look good. Do take a look.

About your concern with redirecting the output, that is completely true. It is how we currently handle it in Wget as well. It is something I have had on my mind, but is not a part of this patch. I'll open another issue, so I don't forget.

darnir:
Do let me know what everyone thinks about these changes. Once we're good to go with these, I can recreate the commits. Post that, my next step would be to create a ring buffer for storing the historical data transfered and hence a speed and ETA calculation

rockdaboot:
Just pulled this branch...
```
bar.c:191:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
 _wget_bar_return_cursor_position() {
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
bar.c: In function ‘_wget_bar_return_cursor_position’:
bar.c:191:1: warning: old-style function definition [-Wold-style-definition]

make[2]: Verzeichnis „/home/tim/src/wget2/src“ wird betreten
  CC       bar.o
  CC       blacklist.o
  CC       job.o
  CC       host.o
bar.c: In function ‘wget_bar_update_thread’:
bar.c:146:14: warning: declaration of ‘bar’ shadows a global declaration [-Wshadow]
  wget_bar_t *bar = (wget_bar_t *) p;
              ^~~
bar.c:57:3: note: shadowed declaration is here
  *bar;
   ^~~
  CC       log.o
  CC       wget.o
  CC       options.o
  CC       utils.o
utils.c: In function ‘determine_screen_width’:
utils.c:45:18: error: storage size of ‘wsz’ isn’t known
   struct winsize wsz;
                  ^~~
utils.c:45:18: warning: unused variable ‘wsz’ [-Wunused-variable]
utils.c:44:7: warning: variable ‘fd’ set but not used [-Wunused-but-set-variable]
   int fd;
       ^~
```

darnir:
Interesting how Clang did not report those. I see them with GCC.

The first two are simple cases. I've pushed a commit that makes the compiler happy about them.

The issue you see in `determine_screen_width` is because you don't seem to to have IOCTL available. Try running `bootstrap` again so gnulib's ioctl module kicks in. 
The reason you see the warnings is because I did not want to litter the code with lots of ifdef's. Hence, I allow the code to declare some variables and then use them inside a `ifdef`. Since in your case, ioctl isn't available, it the variables aren't being used. I think this is a corner case and we can let it be

rockdaboot:
Instead of merging master, is it a problem to rebase onto master ?

rockdaboot:
Why not
```
diff --git a/src/utils.c b/src/utils.c
index 16ed71f..9b3238d 100644
--- a/src/utils.c
+++ b/src/utils.c
@@ -39,19 +39,15 @@
 int
 determine_screen_width (void)
 {
+#ifdef HAVE_IOCTL
   /* If there's a way to get the terminal size using POSIX
      tcgetattr(), somebody please tell me.  */
-  int fd;
   struct winsize wsz;
+  int fd = fileno (stderr);
 
-  fd = fileno (stderr);
-
-#ifdef HAVE_IOCTL
-  if (ioctl (fd, TIOCGWINSZ, &wsz) < 0)
-    return 0;                   /* most likely ENOTTY */
-#else
-  return 0;
+  if (ioctl (fd, TIOCGWINSZ, &wsz) >= 0)
+         return wsz.ws_col;
 #endif
 
-  return wsz.ws_col;
+        return 0;
 }
```

rockdaboot:
Tag indentation *only* works at the beginning of a line (libwget/bar.c, enums).
To fine-tune resp. to set an exact position, you can use spaces after tabs (but never tabs after spaces).

rockdaboot:
Please use xfree() instead of free()

rockdaboot:
IMO, static functions need no prefix, e.g. `_wget_bar_return_cursor_position` could be renamed to `_return_cursor_position`.

rockdaboot:
IMO, in src/bar.c, renaming `wget_bar_update_thread` to `_bar_update_thread` would be consistent.

rockdaboot:
In libwget/bar.c, maybe we can remove 'allocated' flag from _wget_bar_st ?
wget_bar_free() should call wget_bar_deinit().
Then you can replace
```
cleanup:
	free(bar->spaces);
	free(bar->filled);
	free(bar->slots);
	if (allocated)
		free(bar);
```
by
```
cleanup:
	if (allocated)
		wget_bar_free(&bar);
	else
		wget_bar_deinit(bar);
```

rockdaboot:
Feel free to use C99 constructs like
	`for (int it = 0; it < nslots; it++)`
instead of
	`int it; for (it = 0; it < nslots; it++)`
if possible.

rockdaboot:
Use xmalloc and xcalloc

darnir:
Rebased and made changes as suggested

rockdaboot:
Made some changes into a branch with the same name on my repo ... can't make a pull request for you via Github, sorry. Can you see/pull/merge my last changes ?

darnir:
Looks good. I pushed the patch to this branch as well

rockdaboot:
Some things come to my mind...
- did you check with big values for max-threads (exceeding terminal height) ?
- what about error messages ? do you see a chance to print (and scroll) them above the progress area ?

darnir:
Nope. I haven't run any major tests on this yet. But it should not be any different from what was originally there. I just moved it to a separate thread. 

Scrolling terminals allows even larger max-threads values to work just fine.

About error messages, I've been thinking, but couldn't find any good ideas. The best one I can come up with it, on any error, the message should replace the progress bar on that slot. But then it will be replaced once the downloader thread starts downloading something new.printing to the top of the screen is another possibility, but we'll have to set the cursor just right and flush the screen first.

rockdaboot:
Ok, I tested with small terminal height - looks good (or as good as possible :).

Error messages are relatively seldom (lets assume that). We just have to scroll the screen one up, move the cursor to overwrite the first progress line and reprint all the bars below that line. If you could do that, your patch is feature complete (maybe number of bytes have to count up - but I guess that is peanuts !?).

rockdaboot:
Found a way for (error) messages appear above the scroll bars... well, way of.
- scroll the whole screen one up
- print the message in the right line + clear rest of line
- reprint the scroll bar area

No perfect, but works for 'seldom' messages. Maybe you can finalize it !? (Only if you think it's worth doing).

in bar_init() set
```
	// set custom write function for wget_info_printf()
	wget_logger_set_func(wget_get_logger(WGET_LOGGER_ERROR), _info_write);
```

_info_write:
```
static void _info_write(const char *buf, size_t len)
{
//	printf("\033[s\033[1S\033[%dA\033[1G\033[2K", config.num_threads + 2);
	printf("\033[s\033[1S\033[%dA\033[1G\033[0J", config.num_threads + 2);
	fwrite(buf, 1, len, stdout);
	printf("\033[u");
	fflush(stdout);
	for (int i = 0; i < config.num_threads; i++) {
		wget_bar_update(bar, i);
	}
}
```

This loses the red color currently. Maybe we amend log.c/_write_out() !? Not sure...

rockdaboot:
@Darnir Should we go on with that ? WDYT ?

darnir:
Sorry, I've been a little busy and forgot about this in between. 

Your solution looks very god and clean. I still need to test it in some more cases though.
About the loss of colour, maybe we find a way to actually call `_write_error()` from within  `_info_write()`? That would allow us to reuse all the existing code.

darnir:
I've pushed an amended patch that creates a non-static logger function to do this.

Ideally, we'd have some rewrite of the code in log.c to have far more code reuse. If you think this approach is fine, I'll clean up the logging code next.

rockdaboot:
> Ideally, we'd have some rewrite of the code in log.c to have far more code reuse. If you think this approach is fine, I'll clean up the logging code next.

Yes, if you see potential for code cleanup, go ahead.

darnir:
I pushed some refactoring. Does this look good?

rockdaboot:
Does explicitly inlining _write_debug, _write_error, _write_info make sense ?
If you think so, please do it. If you rebase the changes onto master, I'll merge them fast-forward.

darnir:
I think it's better to leave such optimization to the compiler itself. Any modern version of GCC (in the last 5+ years) should be able to automatically detect and flatten these simple method calls.

darnir:
I've rebased all the commits and force pushed to this branch.

rockdaboot:
Ready to merge into master ? Or would you like to merge some commits ?
Maybe brushing up the commit messages to GNU style ?

darnir:
Let me see if I can clean this up a little.

Also, any ideas why the solaris builds are failing?

rockdaboot:
> Also, any ideas why the solaris builds are failing?

Yes, I am in contact with Dagobert Michelsen from OpenCSW about that.
It has nothing to do with your changes - the EUP-JP tests fail due to some (currently) unexplainable character mangling by system(). I hope for a solution the next days.

darnir:
See branch progressbar (https://github.com/darnir/wget2/tree/progressbar) for cleaned-up fast forward ready commits.

rockdaboot:
Could you check/handle the return value of wget_thread_start() in src/bar.c ?

darnir:
Will do

darnir:
Closing in favour of #96

PR #96, opened 2016-09-14
Create an Asynchronous progress bar

darnir:
Give the progress bar its own thread and let it update the entire
display asynchronously at a specified time interval. The existing
implementation refreshes the progress bar for each network packet
downloaded. Over multiple downloader threads and a high speed network
connection this can lead to far too many redrawings of the screen. Also,
each of the downloader threads will block while trying to acquire the
thread mutex because another thread just retrieved a packet. While I
haven't profiled it, it seems like there would be extremely high lock
contention in the existing implementation. Instead, with a separate
thread, we can update all the progress bar slots simultaneously at
regular intervals.

Closes #90. This set of commits is a cleaner version.

### DIFF
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -56,6 +56,7 @@ glob
 iconv
 inline
 inttypes
+ioctl
 lib-symbol-visibility
 listen
 maintainer-makefile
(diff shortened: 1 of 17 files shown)


darnir:
I've rebased the branch against the current master again. You can merge these changes in for now.

The gnulib human module commits can be merged next once we're convinced it works well

PR #97, opened 2016-09-14
Use gnulib human readable

darnir:
I would like to merge these two commits as well. They get rid of the wget_human_readable function and instead use the one provided by gnulib.

### DIFF
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -53,6 +53,7 @@ gettext-h
 gettime
 gitlog-to-changelog
 glob
+human
 iconv
 inline
 inttypes
@@ -110,6 +111,7 @@ sys_stat
 sys_time
 sys_types
 time_r
+unistd
 unlink
 write
 "
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -56,6 +56,7 @@ glob
 iconv
 inline
 inttypes
+ioctl
 lib-symbol-visibility
 listen
 maintainer-makefile
--- a/include/libwget.h.in
+++ b/include/libwget.h.in
@@ -741,6 +741,8 @@ void
 int
 	wget_thread_kill(wget_thread_t thread, int sig) LIBWGET_EXPORT;
 int
+	wget_thread_cancel(wget_thread_t thread) LIBWGET_EXPORT;
+int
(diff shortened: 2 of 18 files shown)


rockdaboot:
Nice. I didn't know that there is such a function within gnulib.
I would like to merge all your commits tomorrow... is that ok ?

rockdaboot:
Any idea why Travis is mourning ?

darnir:
I think I'm fine with the merging.
There is one problem I can currently see, I'll create a new issue for it.

About Travis complaining, I have absolutely no idea. Was hoping you knew more. From a brief glimpse at the logs, it is complaining because of too many errors. The errors all seem to stem from a shadow declaration of the options struct. But I'm not sure how that happens only on Mac and only after my changes. These changes don't affect anything in that area, and the progress bar commits all built fine in #96 .

darnir:
The solaris buildbots are complaining about the same error as Travis on Mac. We really should check this out.

rockdaboot:
Your renaming is no explanation... I'll check on Solaris and report back.

rockdaboot:
It is gnulib 'human' modul that pulls in module 'unistd' which in turn includes the gnulib variant 'getopt.h' via 'unistd.h' - so renaming is a good idea, but maybe we we don't use wget_ as prefix. That is the namespace of libwget, we shouldn't mix that. How do you like 'optionx', 'optionw' or 'woption' ?

Also, we should include the 'unistd' module in bootstrap.conf.

darnir:
Yes, I agree that we should not use `wget_` due to namespace conflicts. This was a test I ran yesterday after looking at the build logs. `optionw` seems like an okay name to me.

What I can't understand is why the human module pulls in these dependencies only on OSX and Solaris. Also, if it does, it should be listed in the modules file, I don't see it there.

rockdaboot:
There is ABOUT-NLS in your commit (that is an auto-generated file).
Where do these changes in po/.gitignore come from ?

darnir:
Aah, those seem to have automatically propped up after I ran a `make maintainer-clean` and rebuilt everything.

darnir:
I have removed the changes from the gitignore files. I don't know where they came from.

rockdaboot:
I just pushed your commits to master. Good work !

Issue #99, opened 2016-09-14
Old downloads disappear from progress output after error_printf()


darnir:
If wget2 generates any error and calls `error_printf()`while progress bar printing is enabled, then a custom logger is used to ensure that the right message is printed on top of the progress bars.

However, since the context of the downloads that have completed has been destroyed, those progress bars are lost and we see blank spaces on the screen.

There are two options to resolve this:

1. Don't destroy the context until either a new one replaces it in the `bar->slot` or the progress bar is destroyed itself
2. Display the messages at the end of the progress bars.

While option 2 looks elegant and simple, there are quite a few issues I see in there. 

1. If the number of errors exceeds the number of rows on the screen, then all calculation will be messed and we'll have garbage on the screen
2. We need to ensure that a new slot will never be added to the progress bars for printing
3. We need to keep track of the number of error messages printed to ensure that we print exactly one below.

rockdaboot:
IMO, we should go with 1. (Don't destroy...). That would also allow a redraw when the screen becomes resized.

PR #102, opened 2016-09-17
Retain a copy of download context for progress bar

darnir:
When the screen is scrolled or resized, we often need to redraw the
entire progress bar. However, when some threads have completed
downloading, this can cause those progress slots to become blank since
their contexts were deleted and hence lost. To prevent this, keep a copy
of the last context in the slot, so we can access it later for printing
the progress bar.

* libwget/bar.c (_bar_slot_t): New member, last_ctx
  (wget_bar_deregister): Create a copy of the context just before we
  deregister. This ensures that we copy the latest data for printing
  (wget_bar_update): If no context is registered, then try printing the
  last context available
  (_bar_print_final): Make the method signature similar to
  wget_bar_update for consistency

### DIFF
--- a/libwget/bar.c
+++ b/libwget/bar.c
@@ -60,7 +60,8 @@ typedef struct {
 	unsigned char
 		first : 1;
 	wget_bar_ctx
-		*ctx;
+		*ctx,
+		last_ctx;
 	char *
 		human_size;
 } _bar_slot_t;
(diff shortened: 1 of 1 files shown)


rockdaboot:
Could you correct the spelling in the last line and then it is fine.

darnir:
Done

darnir:
Where would you like it? Currently it's already printed at the end of the  progress bar.

rockdaboot:
Total # of Bytes in the bottom line. It always shows Bytes: 0 here.

darnir:
That happens because the value of quota isn't being correctly updated. This is because all calls to `quota_modify_read()` use the expression `(config.save_headers ? resp->header->length : 0)`. Since, the --save-headers option isn't passed to Wget, the calls to update quota all add 0 bytes, hence causing the 0 bytes Downloaded in the display as well.

This also implies that we have a bug in the quota compuation

darnir:
This closes #99

rockdaboot:
Good thought, quota fixed in master.

PR #104, opened 2016-09-18
* libwget/net.h: Include netinet/tcp.h to define TCP_FASTOPEN

darnir:


darnir:
P.S. It errored out because savannah is down currently.

rockdaboot:
Sorry, I don't understand this...
We only need netinet/tcp.h in libwget/net.c. And there we have it :-)
It is conditionally included (surrounded by `#ifdef HAVE_NETINET_TCP_H`).

darnir:
You are right. I think I messed up somewhere in my testing which led me to believe that I wasn't using FASTOPEN.

PR #105, opened 2016-09-19
Fix display of total number of bytes downloaded

darnir:
* src/wget.c: Include gnulib module "human".
  (total_download_modify_read): New function for atomically incrementing
  the total_download variable
  (main): Print the human readable version of total_downlaod instead of
  quota when reporting total downloads
  (process_response_part): Update the total_download variable
  (http_recieve_response): Same

### DIFF
--- a/src/wget.c
+++ b/src/wget.c
@@ -46,6 +46,7 @@
 #include <sys/stat.h>
 #include <locale.h>
 #include "timespec.h" // gnulib gettime()
+#include "human.h"
 
 #include "safe-write.h"
 
@@ -126,7 +127,8 @@ static DOWNLOADER
 static void
 	*downloader_thread(void *p);
 static long long
-	quota;
+	quota,
+	total_download;
 static int
 	exit_status,
 	hsts_changed;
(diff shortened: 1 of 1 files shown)


rockdaboot:
Arg, sorry. I missed this. I'll take care for conflict resolving tomorrow...

rockdaboot:
Just pushed the functionality, made it a bit different.
IMO, we should write our own human_readable() - it just introduces lot's of code that we do not need and blows up our binaries. Since we just have a fixed set of options, it should be a few lines of code.
I look at it this afternoon.

darnir:
We did have our own human_readable(). I had originally copied it from GNU Wget. It's shorter, and if you think the gnulib version is adding too much bloat, we can definitely go back to using our own version.

rockdaboot:
Good idea. Just pushed it.
BTW, I found that `valgrind src/wget2_noinstall --progress bar -r www.google.de` spills out invalid reads. Looks like some bar+threading race condition. Maybe you have an idea !?

darnir:
I'll take a look. THe last time I ran the test suite with valgrind, I saw no issues

rockdaboot:
Its not the test suite.

darnir:
Currently, the total number of bytes displayed is incorrect. I haven't investigated it, but using the command line: 
`./src/wget2 --progress bar http://ipv4.download.thinkbroadband.com/5MB.zip http://ipv4.download.thinkbroadband.com/10MB.zip http://ipv4.download.thinkbroadband.com/20MB.zip http://ipv4.download.thinkbroadband.com/50MB.zip`

shows an incorrect total in the summary bar

darnir:
Also, I see no output from valgrind using the command you provided. Can you please share your output?

PR #107, opened 2016-09-25
* Makefile.am: Add targets for checking code coverage

darnir:


rockdaboot:
Code coverage / lcov is so cool ! I never really used it before (just read about it)... and it is such a gain.

darnir:
It is! I set it up long ago for Wget 1.x, but never got anywhere with it. Here, I think we can use this to improve our test coverage and eliminate unneeded functions

rockdaboot:
First thing I saw was 0% test coverage for CSS 😨

darnir:
That and http_highlevel.c surprised me as well. We need quite a few more tests

rockdaboot:
Http_highlevel  ist currently used in examples/ . At some point I would like to run the examples as tests or at least test them somehow.

PR #110, opened 2016-09-27
Support dynamic thread allocation in progress bar

darnir:


src/wget.c
```diff
@@ -839,10 +841,11 @@ static void nop(int sig)
 
 int main(int argc, const char **argv)
 {
-	int n, rc, nthreads = 0;
+	int n, rc;
 	size_t bufsize = 0;
 	char *buf = NULL;
 	char quota_buf[16];
+	nthreads = 0;
```
rockdaboot: Please do not initialize nthreads explicitely - it is already zero.

libwget/bar.c
```diff
@@ -188,8 +190,22 @@ wget_bar_t *wget_bar_init(wget_bar_t *bar, int nslots, int max_width)
...
+		return;
+	/* _bar_print_slot(bar, 0); */
+	fwrite(lf, 1, nslots - bar->nslots, stdout);
+	bar->nslots = nslots;
+	wget_bar_update(bar);
+}
+
 void wget_bar_register(wget_bar_t *bar, wget_bar_ctx *ctx)
 {
+	assert (ctx->slotpos <= bar->nslots);
```
rockdaboot: I don't like assertions in library code. Please check for an error condition and handle it gracefully (e.g. print out an error and avoid stopping/crashing the program).

libwget/bar.c
```diff
@@ -188,8 +190,22 @@ wget_bar_t *wget_bar_init(wget_bar_t *bar, int nslots, int max_width)
 	return NULL;
 }
 
+void wget_bar_set_slots(wget_bar_t *bar, int nslots)
+{
+	char lf[nslots];
+    memset(lf, '\n', sizeof(lf));
```
rockdaboot: Indent with tabs

libwget/bar.c
darnir: This is the reason I added the previous commit to disable assertions. Assertions are disabled by default and enabled only when you enable them explicitly. Also, assertions are never used as a way to stop / crash the program, bu rather they implement checks for the developer. I could completely remove the line here with no issues, since that check isn't mandatory anyhow

rockdaboot: > I could completely remove the line here with no issues, since that check isn't mandatory anyhow

Then it is not needed ;-) Throw it out.

BTW, from `man assert`: "...find bugs ... via a crash that will produce limited debugging output"
So never ever enable assertions - just check with if/else.

darnir: I'll remove and push.

darnir:
Cleaned and pushed. 

I also fixed a memory leak caused by not handling max_threads during bar_deinit

PR #111, opened 2016-09-27
Coverity Fixes

darnir:


darnir:
Thanks. Should I merge it then?

Although the fix about the metalink mirrors seems to be valid. Why do you think it's a false positive?

rockdaboot:
Please don't merge the 'sleep' commit (IMO, that is wrong - but I double check tomorrow).
But merge the metalink one.

> Although the fix about the metalink mirrors seems to be valid. Why do you think it's a false positive?
Because metalink->mirrors always has 1+ entries, else we wouldn't come to that point. That's beyond a static analysis tool. Your code clarifies that, so go ahead. Though we will never be able to 'cover' that code by any tests :-)

rockdaboot:
Arg, you merged the 'sleep' commit... it is only half baken and has a serious bug (mutex must only be unlocked if tcp->caching is true). Also the for loop (tries) does not make sense any more. I'll submit a fix...

darnir:
I merged it before you left the comment. Please fix it if there's a bug. I'll check

rockdaboot:
Just committed

PR #112, opened 2016-09-28
Progress tests

darnir:
I'm not sure if this is the best design, but I quickly whipped this up for some tests. Looks okay to me.

What do you think? This will also give us a little more test coverage.

### DIFF
--- a/libwget/bar.c
+++ b/libwget/bar.c
@@ -226,7 +226,7 @@ void wget_bar_deregister(wget_bar_t *bar, wget_bar_ctx *ctx)
 		last_ctx->raw_downloaded = ctx->raw_downloaded;
 		// Filename will be overwritten when a new file is downloaded by the same
 		// downloader thread. Hence, we make a copy here.
-		last_ctx->filename = strdup(ctx->filename);
+		last_ctx->filename = wget_strdup(ctx->filename);
 	}
 
 	_bar_print_final(bar, ctx->slotpos);
--- a/tests/libtest.c
+++ b/tests/libtest.c
@@ -82,6 +82,8 @@ static char
 static const char
 	*server_hello;
 
+static void wget_test_final(int first_key, const char *switchval, va_list args);
+
 static void sigterm_handler(int sig G_GNUC_WGET_UNUSED)
 {
 	terminate = 1;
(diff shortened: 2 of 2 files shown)


rockdaboot:
I am not so happy with this 'brute force' approach. It slows down valgrind testing a lot for no gain. Most of the tests are very similar and do not influence/change progress code paths.

But let keep this - kind of fuzzing - approach in mind. Extended to all kinds of combinations of options and values, this might find flaws and bugs. It is a different project, though.

For now, we should start with a standard test including --progress. For this we can just copy one of those tests and add '--progress bar'. Also, we should think about unit testing - right now they are in tests.c. Of course this is for libwget testing only.

rockdaboot:
Also, something with the libwget/bar.c code is wrong. Most of the code in wget_bar_init() should go into wget_bar_set_slots(), e.g. the checks `bar->max_slots < nslots`. IMO, wget_bar_init should call wget_bar_set_slots(). WDYT ?

rockdaboot:
Also, we use wget_bar_ctx_t in the library code - but there is no initialization !? wget.c calls bar_register() with an uninitialzed wget_bar_ctx instance. bar_register() gives this to wget_bar_register() which uses the slotpos member as index. What did I miss ?

rockdaboot:
Oh I see. `struct _wget_bar_ctx` is in libwget.h, so it is exposed to anyone. We should not do that... what about wget_bar_register() -> `void wget_bar_register(wget_bar_t *bar, int slot, void *ctx)` ? things like 'filename' are very wget specific, they should not appear the library code. libwget should just hold the abstract API calls, memory (re)allocations and cursor positioning+printing. The content/layout of each 'slot' should be created in src/bar.c because this is a very wget specific thing.

darnir:
Oh I completely agree that this method isn't very great for testing. It's something I quickly whipped up because I wanted to run some very quick tests to see what is working and what isn't. At the end of it, i figured it might be useful to add it to the test suite. 

This test helped me realise that the progress bar code doesn't currently account for some metalink, --delete-after and --chunk-size options. In all these cases, the filename isn't set.

P.S. What you said about the filename being wget specific is absolutely correct. I think I started off with that for a proof of concept and then completely forgot to separate the two out. 

If the content of each slot is made in `src/bar.c` then the entire code for drawing the progress bar needs to move there. Everything in `_bar_update_slot` and `_bar_print_final` should then be in Wget and not libwget. This separation is something I've been thinking about recently.  It also makes it possible to print the filename in the progress bar without requiring libwget to know more about the details of wget.

PR #114, opened 2016-09-30
Progress api

darnir:


darnir:
In general I agree. In fact what you say is similar to what I had in mind. 
Except, init (height = nthreads + 1). We don't need to specify the width, because:

* The libwget/ routines don't need it. Why is the width required there anymore?
* Since each thread prints it's own bar via bar_printf(), then we don't need to allocate the progress buffer of screen_width size within libwget/
* The screen_width may keep changing if the user changes the terminal size. There is no single value and there is no max value either. To keep it completely dynamic, we should not involve the library code in this

The third point is one of the chief reasons, I have been planning to move the progress bar code to client side. It makes resizing for screen width much simpler.

Like you said, the caching stuff should be in src/. Also, you mention that each thread prints its own bar via bar_printf(). If you combine these two statements, we get that the layout of the progress bar is created in src/ and then a `char *` buffer is passed to libwget's wget_bar_print() method to simply output it to the screen at the correct slot.

So, if I take all ideas together, it should work something similar to this:

* Each downloader thread shares a download context with the progress bar routines in src/. This download context contains details like filename, max file size and amount of data downloaded
* The progress bar thread which is created in src/ runs every x milliseconds and for each thread in nthreads, it creates the image of the progress bar using the data shared in the downloader context:
```
for (int i = 0; i < nthreads; i++) {
    wget_bar_print(bar, i, create_progress_image(down_ctx[i]))
}
```
where `create_progress_image(wget_bar_ctx *ctx)` does what `_bar_update_slot()` does currently except that it returns a char * buffer instead of printing it to screen.

I hope I cleared the design well enough by now.

rockdaboot:
Originally I had several tools in mind, all downloading something, all in need of the same progress bar code.
Not repetitions of code, so moved all the code you describe above into a library.
The layout and calculation of width all in the library. The app just sets the data when appropriate.
'data', that is all need to display a bar, e.g. filename, filesize (or -1 if unknown), current bytes downloaded.
That reduces the application code to a minimum, no context needed (just a slot index where we either use downloader.id or we also could use the thread id which then internally maps to a slot number).

The caveat with this approach: The bar layout is hard coded in the library, just with 'fields' that are filled dynamically.

I would use an external library (there exist progress libraries !), if I had found one with multiple lines). All I could find was libraries with one single progress line.

But you are the current progress bar expert and I am really fine with whatever you decide.
My current opinion is: move the 'library bar code' out of the libwget, keep it as separated file in src/. Maybe we later decide to create an own libwget-progress with this code. And just go on as you like. The first goal is to get it working. Thereafter we can optimize/split the code as we want.

darnir:
I agreed with the idea back then. Which is why I created the downloader context which allows the client to fill in the data required for drawing the progress bar while the library does all the work. 

The context is required since only the client code knows how much data has been downloaded for each thread. And the filename and other specific details are also available only from the client. This has the problem, as you mentioned that the design of the progress bar is hard coded. If not, we should remove all that code and put it in src/bar.c . If we want, we can move that to a separate library at a later stage

rockdaboot:
> The context is required since ...

The 'context' could be kept inside the library code. The app just says (based on slot/threadid):
- new download starting, filename=? and expected size=?
- on each read, number of read bytes=?
- end of file: ok, done

Nothing else is needed for a basic progress bar.

> If we want, we can move that to a separate library at a later stage

I am ok with that. Please feel free to do whatever fits best to your design or the amount of time you want to spend.

PR #115, opened 2016-10-03
Refactor progressbar

darnir:
This is a small refactor of the progress bar code based on our discussion on the other pull request. 

While this works fine, it will cause new errors on Coverity. This is because I have now marked some functions in src/bar.c as WGET_NONNULL_ALL. However, currently there still exist a few code paths that do not set the filename and hence, we pass a NULL pointer in such cases.

I'd like to fix those cases soon, but this progress bar should work the same as it currently works.


darnir:
@rockdaboot Do you know why travis fails compilation? I tried a `make disclean` locally and recompiled without any issues. Even `make distcheck` works fine locally for me, yet travis complains.

rockdaboot:
Just seeing
```
bar.c:42:19: fatal error: utils.h: No such file or directory
 #include "utils.h"
                   ^
compilation terminated.
```

That means there is a utils.h in your local path which has not been committed, so a git clone doesn't pull it in.

darnir:
I don't have a local file called "utils.h". Yet local compilation somehow worked for me. I have no idea why.
Anyhow, I have removed the offending include statement, so it should compile well now.

What do you think of the structure of the code now?

rockdaboot:
First I have to earn a coffee... A customer has problems with User-To-User Info on SS7... Hopefully finished soon :smiley:

rockdaboot:
I am fine with your changes. Just see my comments for some details.

include/wget/wget.h
```diff
@@ -265,6 +265,8 @@ int
 	wget_global_get_int(int key) LIBWGET_EXPORT;
 void
 	(*wget_global_get_func(int key))(const char *, size_t) LIBWGET_EXPORT;
+void
+	wget_handle_sigwinch(void) LIBWGET_EXPORT;
```
darnir: Currently, the idea was to provide a library function that acts as a signal handler for SIGWINCH. However, if we rename it to wget_bar_resize, we must ask the client to handle the signals and send libwget just a new width dimension.

darnir: I am not sure if a `wget_bar_resize` style function is a better idea. When the screen size changes. we want the progress bar to automatically adapt to the new screen size. Same thing for if the height changes. If the screen height becomes less that nslots, libwget should refuse to draw the progress bar. The library should not depend on the client giving it the correct information in this case. It should pull the right information on its own.

We can however rename it to wget_bar_screen_changed and retain the current functionality, i.e.: it acts as a signal handler for SIGWINCH or any other method of detecting that the screen size has changed.

rockdaboot: > Currently, the idea was to provide a library function that acts as a signal handler for SIGWINCH.

Yes, but let's assume that libwget does not know about signals (it should be OS independent). Then the app has to fetch the signal and just call wget_bar_resize(void). My idea was that this function is just a 'trigger' function. libwget should do the rest automatically, as you mention.

So basically it stays as it is, but we don't call it *sigwinch*. That name would confuse e.g. a Windows user. Maybe I just was a bit unclear.

darnir: Then I think I understood you just fine. See the commit 354c862. It renames the function as you suggest to make it clearer.

rockdaboot: Almost done... I just cross-compiled your commits - SIGWINCH is unknown :-)))
```
wget.c:839:20: error: 'SIGWINCH' undeclared (first use in this function)
  } else if (sig == SIGWINCH) {
                    ^~~~~~~~
```
Not sure about gnulib... could you fix that and than I'll merge your changes (likely tomorrow morning).

rockdaboot: there is a 'signal-h' modules which provides GNU like signal.h

rockdaboot:
asan fails with clang ?
calculating the screen width doesn't work.

rockdaboot:
Fixed the screen width issue (wget_get_screen_size must not be pure !)

darnir:
Aah. I forgot that it writes to the parameter pointers not just dereference them. Thanks for the fix. 

About the signal-h module, I don't think it provides a SIGWINCH typedef. I'll check again.

rockdaboot:
Maybe not - but there must be some way to check window resizing on Windows (but definitely not via signals). This can be used to get the window size on Windows: http://stackoverflow.com/questions/6812224/getting-terminal-size-in-c-for-windows/12642749#12642749

But for SIGWINCH... simply use an #ifdef

darnir:
I just added an `ifdef` guard in `nop()`. I couldn't find anything useful for this in gnulib.

I think we can add that complexity for Windows support later. If someone is interested in getting this to run on Windows, they will come and help as well

rockdaboot:
I agree.
BTW, I am going to create a make target check-mingw32/64 soon for windows checking. At that point I will implement the code (without resize recognition since windows API doesn't support that).

darnir:
Sure. That sounds like a good idea. Although I think windows cross compilation support can come a little later, after the first *nix only alpha release.

rockdaboot:
We still have the ASAN problem on OSX... any idea ?

```
 Undefined symbols for architecture x86_64:

  "___asan_alloca_poison", referenced from:
```

darnir:
I tried searching. Can't find anything about it online.
Maybe we disable it for now on OSX builds?

Also, I see that on OSX, gcc is just a symlink to clang. In that case, we can eliminate the GCC build on OSX?

rockdaboot:
Yes and yes :-)

darnir:
Quick question: What is the point of `make install` on travis for OSX? It already runs a make distcheck which should catch all the issues

rockdaboot:
Sorry, just remember that sonething went wrong without it. Maybe libwget could not be found during the test run. Give it a try without it.

darnir:
You're right. That was the issue. I've cleaned up the travis scripts and sent a PR

PR #117, opened 2016-10-09
Fix progress http2

darnir:
This pull request also has another unrelated commit. 

Off late, I've been seeing a lot of warnings due to our use of VLAs. These warnings are more informational messages by GCC that VLAs can't be protected. Hence, I decided to silence them.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -173,6 +173,7 @@ if test "$gl_gcc_warnings" = yes; then
   nw=
   nw="$nw -Wsystem-headers"       # System headers may trigger lot's of useless warnings
   nw="$nw -Wvla"                  # This project is C99
+  nw="$nw -Wstack-protector"
 
   # Enable all GCC warnings not in this list.
   gl_MANYWARN_ALL_GCC([warnings])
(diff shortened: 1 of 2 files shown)


rockdaboot:
Wgen exactly do you see vla warnings ? I only see them when having -std=c89 or when using an old c89 compiler. But switching them off is a good thing.

darnir:
It's not vla warnings, but warnings that the stack protector cannot protect
functions that accept vla as a parameter

Thanking You,
Darshit Shah

rockdaboot:
Correction ! I am wrong with 2.
I removed Accept-Encoding from HTTP/2 headers once because the server (google) didn't like it. It has changed since than (the RFC has been finalized meanwhile). I'll fix this now and report back.

rockdaboot:
Just pushed the compression support for http/2.

rockdaboot:
Also just pushed a fix for calculating cur_downloaded. That makes cfb5b0f obsolete.
There still is a issue with redirection:

```
$ src/wget2 --progress bar https://cdimage.debian.org/debian-cd/8.6.0/amd64/iso-cd/debian-8.6.0-amd64-netinst.iso
```

While downloading, 'Bytes' in the last line just showed the # of bytes of the 302 answer.

When finished, it shows (247M is perfect, but 9.98M is unexpected).

```
debian-8.6.0-amd64-n 100% [==========================================>]  247.00M
    [Files: 1  Bytes: 9.98M  Redirects: 1  Todo: 0                            ]
```

Any idea ?

rockdaboot:
> When finished, it shows (247M is perfect, but 9.98M is unexpected).

Just fixed quota calculation and pushed it.

Still the redirection issue has to be solved.

PR #121, opened 2016-11-06
Various enhancements

juaristi:
I find myself in the need of these functions for HPKP, but since they're pretty generic, I put them here, in a separate patch.

 - `wget_base64_get_decoded_length` is a macro that computes the length a base64-decoded string will have after decoding. It basically translates to `((len + 3) / 4) * 3 + 1`, but you don't have to remember it :)
 - Two new functions: `wget_iri_is_ipv4_address` and `wget_iri_is_ipv6_address`, that test whether the host part of an IRI is a literal IPv4/IPv6 address or not. These depend on `inet_pton`, which fortunately is part of gnulib. Actually, I notice the gnulib code is copied verbatim into the [equivalent wget function](http://git.savannah.gnu.org/cgit/wget.git/tree/src/host.c#n445), I don't know why.
 - HSTS should **not** add a host if it's a literal IPv4 or IPv6 address. This is mandated by the spec. The same goes for HPKP as well. And that's why the two functions above come in handy.

### DIFF
--- a/libwget/iri.c
+++ b/libwget/iri.c
@@ -35,6 +35,7 @@
 
 #include <string.h>
 #include <errno.h>
+#include <arpa/inet.h>
 #include "c-ctype.h"
 
 #include <wget.h>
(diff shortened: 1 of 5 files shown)


juaristi:
Yes, good idea. It's done. I've clobbered the old commits. Please have a look now.

libwget/iri.c
```diff
@@ -129,6 +130,30 @@ int wget_iri_isunreserved_path(char c)
...
+
+	if (!host)
+		return 0;
+
+	return inet_pton(AF_INET6, host, &dst);
+#else
+	return 0;
+#endif
+}
+
```
rockdaboot: These function might come handy at other places too. So IMO, they should not belong to the 'iri' namespace.
I guess, they are best placed in 'net' (e.g. wget_net_is_ipv4_address). If in doubt, you could also make them 'private' functions within iri.c. At a later point, if the need arises, we can make them public.
I am not 100% sure about the latter point, so WDYT ?

rockdaboot:
Thanks Ander ! I made some changes to your patch pushed it. 

My changes are: made just one new function wget_ip_is_family() and put it into an own file ip.c, added docs. But functionality is the same.

PR #129, opened 2016-11-16
Migrate from LIBWGET_EXPORT to WGETAPI

### LINKED ISSUE #122: LIBWGET_EXPORT must be preceed the prototypes to work on Windows (MSVC)
Currently we have
```
void
	wget_millisleep(int ms) LIBWGET_EXPORT;
```

but to compile on Windows it must be
```
LIBWGET_EXPORT void
	wget_millisleep(int ms);
```

Currently I added checking for LIBWGET_STATIC to define LIBWGET_EXPORT to be empty.
Thanks to Gisle Vanem for his suggestion.

darnir:
This pull request aims to migrate from using the following form for exporting symbols:

```
void
	wget_millisleep(int ms) LIBWGET_EXPORT;
```

to this form:

```
WGETAPI void
	wget_millisleep(int ms);
```

Here, the attribute for exporting a library symbol now appears *before* the function prototype. Also, the macro for exporting is changed to `WGETAPI` which is more readable and in-line with various other libraries.
Once merged, this should close #122

### DIFF
--- a/include/wget/wget.h
+++ b/include/wget/wget.h
@@ -49,13 +49,13 @@
 
 // see https://www.gnu.org/software/gnulib/manual/html_node/Exported-Symbols-of-Shared-Libraries.html
 #if defined BUILDING_LIBWGET && HAVE_VISIBILITY
-#	define LIBWGET_EXPORT __attribute__ ((__visibility__("default")))
+#	define WGETAPI __attribute__ ((__visibility__("default")))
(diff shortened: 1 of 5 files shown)


darnir:
@Tim: Would it be possible / a good idea to move all the exported functions from tests/libtest.h to `include/wget/wget.h`? That way we can be sure of all the symbols being exported being in one place. 

We could then even `#undef WGETAPI` at the end of the file to prevent exporting symbols from anywhere else.

Issue #132, opened 2016-11-29
Support Environment Variable for config file


darnir:
Wget2 should try to follow the XDG base directory specifications for configuration and data files.

Also, we should have an option to set the user config file via an environment variable similar to `WGETRC`

Also, some options in Wget 1.x are not supported in Wget2 and vice versa, how are we handling these cases?

I was unable to find any information in the manual pages for wget2 about this. I think this is critical information that must be explained before the Alpha release.

rockdaboot:
> Wget2 should try to follow the XDG base directory specifications for configuration and data files.

Not a bug, more of a wishlist. For the alpha release we follow Wget1.x conventions.

> Also, we should have an option to set the user config file via an environment variable similar to WGETRC.

Yes.

> Also, some options in Wget 1.x are not supported in Wget2 and vice versa, how are we handling these cases?

Explained/documented in the [wiki](https://github.com/rockdaboot/wget2/wiki)
We could move that into the docs as .md file.

darnir:
Thanks for adding the last commit, reading $WGETRC and using xdg-dirs. One more thing pending is tilde expansion for filenames, especially hsts-file, ocsp-file and tsl-session. I'm working on writing that using `wordexp`. I'll send a patch soon

rockdaboot:
glob() should be used IMO. It is covered by gnulib... and don't we use it already ?

rockdaboot:
see options.c/get_home_dir(void)

darnir:
`man glob` says this:

> The glob() function searches for all the pathnames matching pattern according to the rules used by the shell (see glob(7)).  No tilde expansion
>        or parameter substitution is done; if you want these, use wordexp(3).

Hence, the choice to use `wordexp`.  I tried setting `hsts-file = ~/.config/wget/hsts-file` in my wgetrc and passed it to wget2. It tried to access a file in a directory called "~", so I guess we are not performing tilde expansion when we read from the config file. The shell does it for us when we use the command line options

darnir:
Okay, reading the man page further shows that the first line in there is wrong. `glob()` can indeed perform tilde expansion when used with `GLOB_TILDE`. Maybe should file a bug with then to change the man page

darnir:
In light of this, I'll change my WIP to use glob instead of wordexp since that gives us better gnulib support as well

darnir:
[...]

PR #134, opened 2016-12-01
Support shell expansion of filenames from wgetrc

darnir:
* bootstrap.conf: Add module wordexp
* src/options.c (_shell_expand): New function to perform local shell
expansion on a given string
(parse_filename): Parse the input string and perform shell expansion on
it using _shell_expand
(parse_string): Set the parser functions for various options that may be
set from a config file to use parse_filename

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -54,6 +54,7 @@
 #include <fcntl.h>
 #include <time.h>
 #include <sys/stat.h>
+#include <wordexp.h>
 //#include <netdb.h>
 
 #include <wget.h>
@@ -246,6 +247,24 @@ static int G_GNUC_WGET_NORETURN print_help(G_GNUC_WGET_UNUSED option_t opt, G_GN
 	exit(0);
 }
 
+static char *_shell_expand(const char *str)
+{
+	char *expanded_str = NULL;
+
+	if (*str == '~') {
+		char *pathptr = strchrnul(str, '/');
+		expanded_str = wget_strnglob(str, pathptr - str, GLOB_TILDE|GLOB_ONLYDIR|GLOB_NOCHECK);
+	}
+
+	// Either the string does not start with a "~", or the glob expansion
+	// failed. In both cases, return the original string back
+	if (!expanded_str) {
+		expanded_str = wget_strdup(str);
+	}
+
+	return expanded_str;
+}
+
 static int parse_integer(option_t opt, const char *val)
 {
 	*((int *)opt->var) = val ? atoi(val) : 0;
@@ -283,6 +302,15 @@ static int parse_numbytes(option_t opt, const char *val)
 			error_printf_exit(_("Invalid byte specifier: %s\n"), val);
 	}
 
+	return 0;
+}
+
+static int parse_filename(option_t opt, const char *val)
+{
+	xfree(*((const char **)opt->var));
+	*((const char **)opt->var) = val ? _shell_expand(val) : NULL;
+
+	debug_printf("Expanded value = %s\n", *(const char **)opt->var);
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


darnir:
This should mostly finish #132. Only thing pending would be documenting these features

rockdaboot:
See my comment  on #132 

With latest gnulib:
```
gnulib-tool: warning: module wordexp doesn't exist
```

rockdaboot:
It's a POSIX function, so at least on Windows we have a problem.

rockdaboot:
```
if (*src == '~')
  expanded_str = wget_str_asprintf("%s%s", get_home_dir(), src + 1);
else
  expanded_str = wget_strdup(src);
```

darnir:
@rockdaboot  I've changed the commit to use `glob()` instead of `wordexp()`. 
This is a WIP commit. I wrote it quickly before class in the morning and there might be possibilities for optimization. This will also cause valgrind to report reachable blocks (No memory leaks).

If you're okay with this, I can also split the function `wget_strnglob` into libwget/utils.c and export it from there. Else, I'll rename it so as not to use the libwget namespace. (Of course, if we move it to utils, even `get_home_dir()`should use it then)

rockdaboot:
@Darshit After reading up a few things and looking at bash's tilde expansion C code, I suggest that we leave tilde expansion to the shell. Each shell has it's special cases when to expand tilde. There is also tilde expansion to $PWD and $OLDPWD.

That means e.g. `--config ~/.wgetrc` works while `--config=~/.wgetrc` doesn't (tilde in latter example won't be expanded). We should document this.

If in doubt, people should use $HOME (or $PWD or whatever) instead of using tilde.

Adding tilde expansion to wget2 would introduce incompatibilities and quirks that we can well do without.

darnir:
@rockdaboot While I in general agree that the tilde expansion should be left to the shell, one very important use case is supporting tilde expansion for files passed from `wgetrc`. Having a hard coded absolute path in the wgetrc should not be a requirement. I don't say we support all forms of tilde expansion, but *atleast* when the command comes from a wgetrc file, we should perform expansion into the $HOME directory. It can be documented that only this feature is available.

This is why I would have liked to use `wordexp` it uses the underlying shell to do the expansions. 
When arguments are passed using the command line, the shell performs whatever expansions are required, so that will always work on all platforms where Wget2 can execute.

However, when reading from a config file, we can document that tilde and shell expansions are only done on supported platforms. Apart from Windows, `wordexp` is available in most recent operating systems. GNULib lists these as platforms with portability issues:
 Mac OS X 10.3, OpenBSD 3.8, Minix 3.1.8, IRIX 5.3, Cygwin 1.5.x, mingw, MSVC 9, Interix 3.5, BeOS. 

They are mostly all 10+ years old. Wasn't one of our goals with Wget2 to target newer platforms and let Wget service the old ones? 

There may be quirks or incompatibilities, but I strongly believe that allowing tilde-filenames inside `wgetrc` is a requirement that we cannot relax.

rockdaboot:
OK. Then we have to document, that we only tilde-expand files from config files and only in a certain way (e.g. point to 'man wordexp' or 'man glob' - whatever you decide for).

So I leave it to you if you prefer wordexp or glob.

But wget_strnglob() curretnly is not reentrant and thus no option for libwget.

darnir:
As much as I'd like to use `wordexp` for its simplicity, I think I'll settle for `glob` mostly because it is far more portable.

A reentrant version of wget_strnglob would use no static variables and we can drop the GLOB_APPEND flag.

src/options.c
```diff
@@ -247,38 +247,18 @@ static int G_GNUC_WGET_NORETURN print_help(G_GNUC_WGET_UNUSED option_t opt, G_GN
...
 {
 	char *expanded_str = NULL;
 
 	if (*str == '~') {
 		char *pathptr = strchr(str, '/');
-		char *homestr = wget_strnglob(str, pathptr - str, GLOB_TILDE|GLOB_ONLYDIR|GLOB_MARK);
-		expanded_str = wget_str_asprintf("%s%s", homestr, pathptr+1);
-		xfree(homestr);
-	} else {
+		expanded_str = wget_strnglob(str, pathptr - str, GLOB_TILDE|GLOB_ONLYDIR|GLOB_NOCHECK);
```
darnir: You are right, I missed that. I've pushed a fix. Once everything looks good, I will squash, write a clean commit message and push

rockdaboot: Maybe you can squash it right now, makes review easier on Github.
I'll have a closer look at it tomorrow...

rockdaboot:
I am fine with your patch, just add 'strchrnul' to bootstrap.conf.

rockdaboot:
Any progress here ? strchrnul meanwhile is in bootstrap.conf, src/options.c needs small changes.

darnir:
I am on vacation for a week. Needs a very small set of changes which I'll
finish by the end of the week.

Thanking You,
Darshit Shah

rockdaboot:
Thanks Darshit. Slightly amended and pushed.

PR #150, opened 2017-03-22
WIP : Implementing --no-proxy to not use proxy

### LINKED ISSUE #149: Implement --no-proxy to not use proxy even if proxy environment variables set
Also we need to implement  `--proxy` option to use proxy when proxy environment variables set. Should this be filed under separate issue?

rootkea:
This is an almost complete attempt to fix #149 

`wget2` now works with `--proxy` option.
`wget2` also works without `--proxy` option but proxy been set in environment (default behavior)

Last thing remaining is to check for the inverted value of `proxy` in the case of `--no-proxy` as currently even with `--no-proxy` wget2 fetches web-page over proxy.

This commit doesn't include tests. I'll write them once I finish the last bit.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -1445,7 +1445,7 @@ int init(int argc, const char **argv)
 	debug_printf("Local URI encoding = '%s'\n", config.local_encoding);
 	debug_printf("Input URI encoding = '%s'\n", config.input_encoding);
 
-	if(config.proxy) {                                                     // This needs inspection!
+	if (config.proxy) {                                                     // This needs inspection!
 		char *ptr;
 		if ((ptr = getenv("http_proxy")) && wget_http_set_http_proxy(ptr, config.local_encoding) < 0) {
 			error_printf(_("Failed to set environment http proxies %s\n"), ptr);


src/options.c
```diff
@@ -1443,6 +1445,18 @@ int init(int argc, const char **argv)
 	debug_printf("Local URI encoding = '%s'\n", config.local_encoding);
 	debug_printf("Input URI encoding = '%s'\n", config.input_encoding);
 
+	if (config.proxy) {                                                     // This needs inspection!
```
darnir: Prefer keeping comments on their own line. It makes them easier to read

rootkea: Ok. Thanks. From next time I'll adhere to comments on separate line policy :)

darnir:
Seems fine to me in general.
When `--no-proxy` is defined, we should also ignore any `--http-proxy` and `--https-proxy` directives. So in case of the inverse, you should explicitly unset any proxy settings. This is useful because a user may set `http-proxy` in their `wget2rc` file which should also be overridden using the `--no-proxy` switch

rootkea:
@darnir I thought `--no-proxy` is used to ignore environment proxies and not the option proxies. 

    $ man wget
    --no-proxy
           Don't use proxies, even if the appropriate *_proxy environment variable is defined.

Also even if we extend `--no-proxy` to disable option proxies i.e. `--http-proxy` and `--https-proxy` what is the expected behavior when these option proxies come after `--no-proxy`? i.e.

    wget2 --no-proxy --http-proxy=http://192.168.8.253:3128

rockdaboot:
I just took a look at wget1.x and found out that --no-proxy is not what's written in the docs.
The docs say:
```
--no-proxy
  Don't use proxies, even if the appropriate *_proxy environment variable is defined.
```

The code want's to have a comma-separated list of proxies as argument. This is taken in addition to the list given by the environment variable 'no_proxy'. If the argument is empty (or not given ?), 'no_proxy' env var is ignored and the list stays empty.

To be backward compatible, we should mimic a similar behavior.

Sorry, but I was misguided by 'man wget' and it's description of --no-proxy.

rootkea:
> Sorry, but I was misguided by 'man wget' and it's description of --no-proxy.
Same here. :)

rootkea:
But now after this changed definition of `--no-proxy` I don't see a point in having `--proxy` option. As by default anyhow environment proxies are going to be honored by `wget2`. Please correct me if I'm wrong.

darnir:
@rockdaboot I'm not sure maintaining full compatibility in such a case is a very good idea. I would vote to break 1.x compatibility in this case for a more intuitive solution where `--no-proxy` simply disables all proxy settings. Is there any usecase where this would cause major problems?

rootkea:
@darnir But let's not forget that traditionally, the environment variable `no_proxy` has been used to provide urls for which proxy shouldn't be used and not as a flag to discard proxy for *all* the urls.

rockdaboot:
Well --no-proxy and --proxy seems to be used in the real world. You'll see when you search the internet for 'wget --no-proxy'. I have not much time any more today, but from what I saw on a first glimpse, people use
```
--proxy=<proxy> to set a proxy
--no-proxy to switch proxy off
--no-proxy=<list of CIDRs> to disable proxying just for the given addresses
```

But please check for your own, I am online tomorrow morning.

darnir:
> ```
> --proxy=<proxy> to set a proxy
> --no-proxy to switch proxy off
> --no-proxy=<list of CIDRs> to disable proxying just for the given addresses
> ```

This looks like a good way to implement it. An empty string implies don't use proxies anywhere.
Of course, then --no-proxy needs to be explicitly defined as an option and the --proxy switch is simply a shorthand for --http-proxy and --https-proxy

rootkea:
Since we need to dig deep about `--no-proxy` let's talk about `--proxy` and let's get this over with.

What's the use of `--proxy` option?

1. To set proxies? 
No. You use `--http-proxy` and `--https-proxy` options for that.

2. To ask `wget2` to use environment proxies?
Again, since it's a default intended behavior why should we have `--proxy` option for something which `wget2` is supposed to do by default?

rootkea:
Is there any update on this? I still can't understand why do we need `--proxy` option

rockdaboot:
As @darnir said `the --proxy switch is simply a shorthand for --http-proxy and --https-proxy`.

rootkea:
Ok. Revisiting this one as right now there are 3 issues which concern with "proxy". Here I'm consolidating all the discussion that has happened until now.

1. `wget2` by default uses proxy environment variables.
2. When supplied with `--http-proxy=<addr>, `https-proxy=<addr>` and `--proxy=<addr>` new proxies are set overwriting the one we inherited from environment.
3. When supplied with `--no-proxy=<addr, addr>` it disables proxies for these specific hosts
4. If supplied with `--no-proxy` it disables the proxies completely.

BTW if 2) is true then we need to change this line from `wget2 --help`
> --proxy             Enable support for *_proxy environment variables. (default: on)

Are the above conclusions correct? Just making it clear so that this sets the ground for further work.

Also I'm sure option parsing code does the job of reflecting the last option on command line when conflicting options are passed e.g. `wget2 --no-proxy --http-proxy=http://192.168.8.253:3128`

rootkea:
Can anybody please confirm or refute the conclusions drawn in above comment? Thanks.

darnir:
That seems like a good workflow and how I would expect the proxy switches to work. Also, you're correct, the option parsing code will pretty much handle the conflicting options based on the order they're passed. This is important since, someone may add `no-proxy` to the configuration file and then pass a `--proxy=<addr>` switch on the command line.

PR #151, opened 2017-03-22
free libidn memory using idn_free() on Windows platform

### LINKED ISSUE #144: Misuse of libidn memory
An old [issue](http://lists.gnu.org/archive/html/bug-wget/2015-02/msg00035.html) from Wget that has crept across to Wget2:

In `libwget/iri.c` and the function `iri_free_content()`, the `iri->host` memory-block can come from `idna_to_ascii_8z()`  (depending on how Wget2 was configured).

In other words, if using IDNA as a DLL on Windows, this memory could have been allocated by another CRT. So `idna_free()` should be called to free this block. 

In MSVC-2015 and using a MinGW compiled `libidn.dll.a`, this issue is clear since different CRTs are used; `MSVCRT.DLL` vs `VCRUNTIME140.DLL+UCRTBASE.DLL`. Callstack from `test.exe`:
```
ntdll!RtlpFreeHeap+0x7b5
ntdll!RtlFreeHeap+0x758
ucrtbase!_free_base+0x1b
ucrtbase!free+0x18
libwget2!wget_iri_free_content(struct wget_iri_st * iri = 0x02bcdf88)+0x22
test!invoke_main+0x1d
test!__scrt_common_main_seh(void)+0xff
...
```

Maybe the same issue for IDN2 (I have no experience with it).

rootkea:
Attempt to fix #144 

I will add tests once code gets approved.

I don't have Windows machine so please do comment whether this fixes the issue or not.

### DIFF
--- a/libwget/encoding.c
+++ b/libwget/encoding.c
@@ -227,7 +227,7 @@ const char *wget_str_to_ascii(const char *src)
 	if (wget_str_needs_encoding(src)) {
 		char *asc = NULL;
 		int rc;
-#if defined WITH_LIBUNISTRING && IDN2_VERSION_NUMBER < 0x00140000
+# if defined WITH_LIBUNISTRING && IDN2_VERSION_NUMBER < 0x00140000
 		uint8_t *lower, resbuf[256];
 		size_t len = sizeof(resbuf) - 1; // leave space for additional \0 byte
 
@@ -255,17 +255,23 @@ const char *wget_str_to_ascii(const char *src)
 
 		if (lower != resbuf)
 			xfree(lower);
-#else
-#if IDN2_VERSION_NUMBER < 0x00140000
-		if ((rc = idn2_lookup_u8((uint8_t *)src, (uint8_t **)&asc, 0)) == IDN2_OK) {
-#else
-		if ((rc = idn2_lookup_u8((uint8_t *)src, (uint8_t **)&asc, IDN2_NONTRANSITIONAL)) == IDN2_OK) {
-#endif
+# else
+#  if IDN2_VERSION_NUMBER < 0x00140000
+		if ((rc = idn2_lookup_u8((uint8_t *)src, (uint8_t **)&asc, 0)) == IDN2_OK)
+#  else
+		if ((rc = idn2_lookup_u8((uint8_t *)src, (uint8_t **)&asc, IDN2_NONTRANSITIONAL)) == IDN2_OK)
+#  endif
+		{
 			debug_printf("idn2 '%s' -> '%s'\n", src, asc);
-			src = asc;
+#  ifdef _WIN32
+				src = wget_strdup(asc);
+				idn2_free(asc);
+#  else
+				src = asc;
+#  endif
 		} else
 			error_printf(_("toASCII(%s) failed (%d): %s\n"), src, rc, idn2_strerror(rc));
-#endif
+# endif
 	}
 #elif WITH_LIBIDN
 	if (wget_str_needs_encoding(src)) {
@@ -277,7 +283,12 @@ const char *wget_str_to_ascii(const char *src)
 
 			if ((rc = idna_to_ascii_8z(src, &asc, IDNA_USE_STD3_ASCII_RULES)) == IDNA_SUCCESS) {
 				// debug_printf("toASCII '%s' -> '%s'\n", src, asc);
+# ifdef _WIN32
+				src = wget_strdup(asc);
+				idn_free(asc);
+# else
 				src = asc;
+# endif
 			} else
 				error_printf(_("toASCII failed (%d): %s\n"), rc, idna_strerror(rc));
 		}


rootkea:
Oh forgot about `xfree(iri->host)` from `iri.c`. If on Windows then we have already freed the `libidn` memory. Need to inspect this closer.

gvanem:
How is your change going to return a valid `src`?

rootkea:
@gvanem  I think I need to allocate memory for `src`. Just let me look into code more if that's the case indeed.

rootkea:
What do you think about this?

    #ifdef _WIN32
				if (!(src = malloc((strlen(asc) + 1) * sizeof(char)))) {
					printf("malloc failed.\n");
					exit(1);
				}

				strncpy(src, asc, sizeof(asc) + 1);
				idn_free(asc);
    #else
				src = asc;
    #endif

The part which I don't get is what about the earlier value `src` was pointing at? The one we received as an argument? When we say `src = asc` or `src = malloc()`, that initial reference is lost. How do we free that resource? This also applies to original code. Line no. 280 encoding.c

Also I just noticed we also need to inspect libidn2. More specifically `idn2_lookup_u8()`.

rootkea:
Oh I just noticed we can also use `wget_strdup` instead of `malloc` and `strncpy`. I have no idea if using `wget_strdup` is more efficient aside from the fact that it's just a single function call. (But it might be calling `malloc` and `strcpy` under the hood)

rockdaboot:
wget_malloc: (in Netbeans: right click, navigate->declaration and you see what it is doing).
It never returns NULL, so no check needed. In a out-of-memory situation it calls a function that is customizable and exits. You can also read the docs for this function.

wget_strdup is also documented. It returns NULL if you give it a NULL pointer, else it works like strdup(), but is more portable. Since it uses wget_malloc() internally, same for OOM as above.

BTW, never use strncpy(). It has side-effects and sometimes doesn't 0-terminate a string. On *BSD* the linker gives you are warning if you use it :-)

rockdaboot:
Regarding your code: Right idea, just use our (standard) functions, like this
```
#ifdef _WIN32
	src = wget_strdup(asc);
	idn_free(asc);
#else
	src = asc;
#endif
```

Please also do that corresponding for both calls of idn2_lookup_u8(), just using idn2_free() instead of idn_free().

rootkea:
@rockdaboot Thanks for the explanation. :)

On a funny note, initially they were like "don't use `strcpy()`" because (quoting manpage)

> If  the  destination  string of a strcpy() is not large enough, then anything might happen.  Overflowing fixed-length string buffers is a favorite cracker technique for taking complete control of the machine.

They said use `strncpy()` instead to control the `n` and hence to avoid buffer overrun.

But now `strncpy()` doesn't null terminate the string if `\0` is not in those `n` characters thereby opening a whole new battleground!

It seems like `strdup()` is the *only* sane way to copy strings. In our case it's `wget_strdup()`

darnir:
That is why the BSD's introduced `strfcpy()`.

rockdaboot:
strlcpy() ? We have that function in libwget too :-)

It is technically a shortcut for `snprintf(buf, bufsize, "%s", s)`.

rockdaboot:
> It seems like strdup() is the only sane way to copy strings.

There are many ways to copy strings. It depends on the use case.
strncpy() is badly designed, but that is not the problem. The problem are people not knowing or caring about possible side-effects though they are clearly documented.

strlcpy() also has it's side effects that you should be aware of. Think about the return value... yes, it returns the length of the `src`. Sometimes your src is *much* longer than the dst, but you are only interested in the first x bytes and do not care about how long src is. This can introduce many wasted CPU cycles for nothing. In fact, I head this during the wget2 development once and it took me some time to figure out what ate up so many CPU cycles.

rockdaboot:
My conclusion: now your tools exactly, else you fail in the long term.

rootkea:
I took some liberty to remove two opening parenthesis after if statements (line no. 260 and 262) and added one single opening parenthesis on a new line as the concerned function has  one less closing parenthesis than total no. of opening parenthesis (but it still works because part of code gets removed by preprocessor directives.)

I also formatted the #if, #else, #ifdef etc. directives as the code was turning into spaghetti!

rootkea:
I know I deviated from the coding style we use i.e. to have opening parenthesis on a same line but I did that to maintain the sanity while looking at the code and wondering where's that one lost closing parenthesis. If it's not okay to deviate I'll update the code. :)

rockdaboot:
The coding style is not set in stone and readability is a major concern. In this case, well, your deviation is ok (not giving me personally any pros or cons).

Before we can merge, please merge the 3 commits into one (git rebase -i HEAD~3) and write a proper commit message (GNU style, see git log for examples). Then `git push -f`.

rootkea:
I did `git push -f`
But then it says:

    fatal: The current branch libidn has no upstream branch.
    To push the current branch and set the remote as upstream, use

    git push --set-upstream origin libidn

This will push to my own repo on github, right? As `git remote -v` shows

    origin	https://github.com/rootkea/wget2.git (fetch)
    origin	https://github.com/rootkea/wget2.git (push)
    upstream	https://github.com/rockdaboot/wget2.git (fetch)
    upstream	https://github.com/rockdaboot/wget2.git (push)

So then after doing above command should I send another pull request?
Sorry I'm asking you about git but I'm really confused. I read about `push`, `pull`, `merge` etc. But I just don't have a proper workflow.

rootkea:
Ok. I messed things up. I now see 3 another commits with same changes. I really feel helpless when it's come to git. I need to experiment things with a dummy repo, it seems.

darnir:
It's okay. Usually you should not do a force push unless you're sure of what you are doing. It can cause loss of data.

I see a total of 6 commits in your repository on this branch. 

```
*   e01868e [N] - (HEAD -> libidn, rootkea/libidn) Merge branch 'libidn' of https://github.com/rootkea/wget2 into libidn (8 minutes ago) <Avinash Sonawane>
|\
| * 08109b1 [N] - free libidn and libidn2 memory using corresponding functions (2 hours ago) <Avinash Sonawane>
| * 7b6911d [N] - Fixed formating by adding space between `while` and `(` (19 hours ago) <Avinash Sonawane>
| * a142988 [N] - free libidn memory with idn_free() on Windows platform (19 hours ago) <Avinash Sonawane>
* | 1f87b06 [N] - Fix libidn & libidn2 memory misuse (25 minutes ago) <Avinash Sonawane>
* | 7fd33bc [N] - Fixed formatting by separating `while` and `(` (35 minutes ago) <Avinash Sonawane>
* | 8afc11b [N] - Fix misuse of libidn & libidn2 memory (41 minutes ago) <Avinash Sonawane>
|/

```

Did you mean to add more commits?

In the future, what you can do is keep pushing new commits to the same branch. Don't try to force push a new commit. Once the review is over, you can refactor the commits into something more meaningful.

darnir:
Okay.. As I see it, you somehow managed to mess up the git history by having different commit messages. I'm not entirely sure of how it may have transpired. One way to fix this mess without going too deep into Git is the following set of commands:

```
$ git checkout master
$ git checkout -b libidn_new
$ git merge --squash libidn
$ git commit
$ git format-patch -1
$ git checkout libidn
$ git reset --hard origin/master
$ git am <name of patch file generated>
$ git push -f origin libidn
```

There are other ways to achieve this with fewer commands (`git rebase`), but those commands always leave the possibility of something being messed up. Using simpler but more commands is a safe option to choose.

libwget/encoding.c
```diff
@@ -277,7 +283,12 @@ const char *wget_str_to_ascii(const char *src)
 
 			if ((rc = idna_to_ascii_8z(src, &asc, IDNA_USE_STD3_ASCII_RULES)) == IDNA_SUCCESS) {
 				// debug_printf("toASCII '%s' -> '%s'\n", src, asc);
+# ifdef _WIN32
+				src = wget_strdup(asc);
+				idn_free(asc);
```
gvanem: This call is not proto-typed anywhere. Hence:
```
# include <idn-free.h>`
```
is needed at the top. I assume if one has `<idna.h>` one has this header too.

rootkea:
@darnir Thank you! Those git operations did solve the issue. But the reason why I did what I did was because I wanted to merge the 3 commits into one (git rebase -i HEAD~3) and write a proper commit message as @rockdaboot suggested. And now I can see one commit which squashed all other commits with no proper message as Tim suggested. :(

And I can't see a way to edit the commit message. Need to read up more on that.

darnir:
You should have edited the message when you ran `git commit`. Anyways, you can now use:
`$ git commit --amend`. It should open up an editor where you can edit the commit message. Once done, do another force push.

libwget/encoding.c
darnir: Interesting. I see that even on my system, I have the `idn-free.h` header, but the man page for`idn_free()` states that it requires only `<idn.h>`.

rockdaboot:
Thanks, commit message amended and merged into master.

rockdaboot:
Just for practice (merging last 3 commits and amend commit message):
- Check where you are with `git branch` and/or with `gitk` at will in between

```
$ git checkout master # make sure you are on master
$ git checkout -b test # create 'test' from master and switch to it
$ git rebase -i HEAD~3 # now vi opens with last 3 commits in 3 lines
# edit the lower 2 lines: exchange 'pick' with 'f'
# now save & quit
$ gitk # just to see your commits and the content, all 3 commits merged now
$ git commit --amend # opens editor, edit the commit message and save&quit
$ gitk # just to see that your message has been changed
# Now you could `git push -f`, but as @darnir says, do that only for private branches (or if I say so ;-))
$ git checkout master # switch to master
$ git branch -D test # cleanup: delete branch test
```

All the changes were just 'locally' to your 'test' branch, nothing else changed.

git takes a while before you feel save - but then you will love it for sure !

libwget/encoding.c
rockdaboot: Explanation is in idn-free.h (it is just made for Windows). Just pushed a patch to include it on _WIN32.

rockdaboot:
Please take some time and read 'Getting Started' and 'Git Basics' to get an idea of how it works: http://git-scm.com/doc

When it comes to branching, here is a good article (also explains some git basics): http://nvie.com/posts/a-successful-git-branching-model/ 

You will certainly have to use Git almost everywhere nowadays, to take your time with it and learn it. It is as valuable as knowing compiler options or C syntax (also when you go for a job).

rootkea:
@rockdaboot Thank you! Tonight's task is to take on git. Then tomorrow morning I'll start working on my proposal. I better hurry!

PR #152, opened 2017-03-23
Set some noreturn, const and pure attributes in Clang as well

darnir:
* include/wget/wget.h: Set function attributes noreturn, const and pure
when using the Clang compiler as well.

### DIFF
--- a/include/wget/wget.h
+++ b/include/wget/wget.h
@@ -74,7 +74,7 @@
 #	define GCC_VERSION_AT_LEAST(major, minor) 0
 #endif
 
-#if GCC_VERSION_AT_LEAST(2,5)
+#if GCC_VERSION_AT_LEAST(2,5) || defined(__clang__)
 #	define G_GNUC_WGET_CONST __attribute__ ((const))
 #	define G_GNUC_WGET_NORETURN __attribute__ ((noreturn))
 #else
@@ -90,7 +90,7 @@
 #	define G_GNUC_WGET_UNUSED
 #endif
 
-#if GCC_VERSION_AT_LEAST(2,96)
+#if GCC_VERSION_AT_LEAST(2,96) || defined(__clang__)
 #	define G_GNUC_WGET_PURE __attribute__ ((pure))
 #else
 #	define G_GNUC_WGET_PURE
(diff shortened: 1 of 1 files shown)


rockdaboot:
Here, clang always mimics to be gcc 4.2. So your patch is not needed:
```
$ clang -dM -E - < /dev/null|grep GNUC
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
```

This is true for clang-3.6 and even clang-5.0.
What am I missing ?

darnir:
Well, to be honest, I didn't run such a check at all. My reasoning for this patch was @gvanem's comment  here https://github.com/rockdaboot/wget2/commit/18dfa9f39e4ded9dd7bbb76c965653608545ae32#commitcomment-21456830

The reason the compiler complained about an uninitialized variable was because it didn't realise that `_no_memory` would not return. This is despite the fact that it has been tagged as such. As a result, I deduce that Clang did not pass the `GCC_VERSION_ATLEAST(2,5)` test.

rockdaboot:
Maybe he can tell us his clang version... (wrote a comment to him)

gvanem:
@rockdaboot  Which version is your clang ? Normally they mimic gcc 4.2 and should understand the 
@rockdaboot noreturn attribute of gcc... so it's either a bug in clang or you have an outdated version !?

On the contrary, I use 3.9:
```
c:\>clang-cl -v
clang version 3.9.0 (branches/release_39)
Target: i686-pc-windows-msvc
Thread model: posix
```

But note, it's `clang-cl`, the one that mimics MSVC. Since that's the only way AFAICS there is to compile Wget2 for MSVC. Since MSVC-2015 is still not C99-compatible enough.

rockdaboot:
There is a clang-cl on Linux as well (e.g. /usr/lib/llvm-3.9/bin/clang-cl) which mimics MSVC. But since the environment is totally different, it can't even find standard includes.

Anyways, we do more checks with Linux clang on the CI. Also, @darnir, you can't test if your code works out with clang-cl and we know the code is fine.

If @gvanem wants clang-cl being silent, he should open a (clang-cl tested) pull request - he is currently the only person that can test such a change.

PR #154, opened 2017-03-25
Fix for separate build trees

akashrawal:
Temporary fix for [separate build trees](https://www.gnu.org/software/automake/manual/automake.html#VPATH-Builds).

A more "proper" fix might be to place generated files in a separate directory.


rockdaboot:
Thanks, merged via command line (amended commit messages).

PR #155, opened 2017-03-27
* tests/test.c: Add wget_robots_parse unit test

dstw:
This is my GSoC 2017 microproject. To get more familiar with Wget2 codebase, with guide from Tim Ruehsen [1], I tried to create a simple unit test which call wget_robots_parse() from libwget to process data that contains robots.txt information.

[1] http://lists.gnu.org/archive/html/bug-wget/2017-03/msg00101.html

### DIFF
--- a/tests/test.c
+++ b/tests/test.c
@@ -1992,6 +1992,155 @@ static void test_netrc(void)
 	rmdir(".test");
 }
 
+static void test_robots(void)
+{
+	static const struct test_data {
+		const char *
+			data;
+		const char *
+			path[3];
+		const char *
+			sitemap[3];
+	} test_data[] = {
+		{
+			// Deny all robots from part of the server
+			"User-agent: *\n"
+			"Disallow: /cgi-bin/\n",
+			{ "/cgi-bin/", NULL },
+			{ "", NULL }
+		},
+		{
+			// Deny all robots from the entire server
+			"User-agent: *\n"
+			"Disallow: /\n",
+			{ "/", NULL },
+			{ NULL }
+		},
+		{
+			// allow a single robot for part of the server
+			"User-agent: wget2\n"
+			"Disallow: /cgi-bin/\n",
+			{ "/cgi-bin/", NULL },
+			{ NULL }
+		},
+		{
+			// deny a single robot
+			"User-agent: wget2\n"
+			"Disallow: /\n",
+			{ "/", NULL },
+			{ NULL }
+		},
+		{
+			// allow a single robot
+			"User-agent: *\n"
+			"Disallow: /\n"
+			"User-agent: wget2\n"
+			"Disallow: \n",
+			{ "/", NULL },
+			{ NULL }
+		},
+		{
+			// Allow all robots complete access
+			// or simply don't use a robots.txt at all
+			"User-agent: *\n"
+			"Disallow: \n",
+			{ NULL },
+			{ NULL }
+		},
+		{
+			// with 1 sitemap
+			"User-agent: *\n"
+			"Disallow: /cgi-bin/\n"
+			"Sitemap: http://www.example.com/sitemap.xml",
+			{ "/cgi-bin/", NULL },
+			{ "http://www.example.com/sitemap.xml", NULL }
+		},
+		{
+			// with 2 sitemap
+			"User-agent: *\n"
+			"Disallow: /cgi-bin/\n"
+			"Sitemap: http://www.example1.com/sitemap.xml\n"
+			"Sitemap: http://www.example2.com/sitemap.xml",
+			{ "/cgi-bin/", NULL },
+			{ "http://www.example1.com/sitemap.xml", "http://www.example2.com/sitemap.xml", NULL }
+		},
+		{
+			// with 2 "Disallow" entries
+			"User-agent: *\n"
+			"Disallow: /cgi-bin/\n"
+			"Disallow: /tmp/\n"
+			"Sitemap: http://www.example1.com/sitemap.xml\n"
+			"Sitemap: http://www.example2.com/sitemap.xml",
+			{ "/cgi-bin/", "/tmp/", NULL },
+			{ "http://www.example1.com/sitemap.xml", "http://www.example2.com/sitemap.xml", NULL }
+		},
+		{
+			// with equal number entries
+			"User-agent: *\n"
+			"Disallow: /cgi-bin/\n"
+			"Disallow: /tmp/\n"
+			"Disallow: /junk/\n"
+			"Sitemap: http://www.example1.com/sitemap.xml\n"
+			"Sitemap: http://www.example2.com/sitemap.xml\n"
+			"Sitemap: http://www.example3.com/sitemap.xml",
+			{ "/cgi-bin/", "/tmp/", "/junk/" },
+			{ "http://www.example1.com/sitemap.xml", "http://www.example2.com/sitemap.xml", "http://www.example3.com/sitemap.xml" }
+		},
+		{
+			// null termination test
+			"User-agent: *\n"
+			"Disallow: /cgi-bin/",
+			{ "/cgi-bin/", NULL },
+			{ "", NULL }
+		},
+		{
+			// null termination test #2
+			"User-agent: *"
+			"Disallow: /cgi-bin/",
+			{ "/cgi-bin/", NULL },
+			{ "", NULL }
+		}
+	};
+
+	for (unsigned it = 0; it < countof(test_data); it++) {
+		const struct test_data *t = &test_data[it];
+		ROBOTS *robots = wget_robots_parse(t->data, PACKAGE_NAME);
+
+		for (unsigned it2 = 0; it2 < countof(test_data[it].path) && t->path[it2]; it2++) {
+			int n = wget_vector_size(robots->paths);
+			for (int it3 = 0; it3 < n; it3++) {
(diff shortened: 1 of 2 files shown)


tests/test.c
```diff
@@ -1990,6 +1990,44 @@ static void test_netrc(void)
 	rmdir(".test");
 }
 
+static void test_robots(void)
+{
+	ROBOTS *robots;
+	ROBOTS_PATH *path;
+	const char *sitemap;
+	const char *data =
```
rockdaboot: You should declare a test_data structure with input (what now is 'data') and with expected output (e.g. array of paths, array of sitemaps). For example, see one function further up - test_netrc().

tests/test.c
```diff
@@ -1990,6 +1990,44 @@ static void test_netrc(void)
...
+					"Disallow: /maps/reserve/b/\n"\
+					"Disallow: /about/views/\n"\
+					"Disallow: /intl/*/about/views/\n"\
+					"Sitemap: http://www.gstatic.com/s2/sitemaps/profiles-sitemap.xml\n"\
+					"Sitemap: https://www.google.com/sitemap.xml";
+
+	robots = wget_robots_parse(data, PACKAGE_NAME);
+
+	for (int it = 0; it < wget_vector_size(robots->paths); it++) {
+		if (!(path = wget_vector_get(robots->paths, it))) {
```
rockdaboot: This is always true, you just iterate over all found paths. You want to compare them with an expected array/list of paths.

tests/test.c
```diff
@@ -1990,6 +1990,44 @@ static void test_netrc(void)
...
+			info_printf("Failed to get path '%s'\n from robots", path->path);
+			failed++;
+		} else {
+			ok++;
+		}
+	}
+
+	for (int it = 0; it < wget_vector_size(robots->sitemaps); it++) {
+		if (!(sitemap = wget_vector_get(robots->sitemaps, it))) {
```
rockdaboot: This is always true, you just iterate over all found sitemaps. You want to compare them with an expected array/list of sitemaps.

dstw:
I have create data with random carriage return position (just like in function test_netrc) and compare this with hard coded value. Btw, how could I check the coverage percentage for just robots.c function?

rockdaboot:
> Btw, how could I check the coverage percentage for just robots.c function?

After `make check-coverage` just view the results as HTML, e.g.:
`firefox lcov/index.html`

Click on the /libwget directory and you'll see each file listed with coverage numbers.

dstw:
Based on lcov report, I still found 4 more lines are still miss. Will work on this.

rockdaboot:
> Based on lcov report, I still found 4 more lines are still miss. Will work on this.

Thats great !
BTW, have a look why Travis CI fails and try to fix it or ask if you are lost.

And a (yet) hidden feature for developers: `touch .manywarnings` and do another build `./configure && make clean && make && make check`.
This turns on all kinds of gcc warnings - if you see one, try to eliminate it.

rootkea:
@rockdaboot After turning many warning on `make` fails.

    gcc: error: unrecognized command line option '-Warray-bounds=2'
    gcc: error: unrecognized command line option '-Wshift-overflow=2'
    gcc: error: unrecognized command line option '-Wunused-const-variable=2'

    $ gcc --version
    gcc (Debian 4.9.2-10) 4.9.2

After checking further it looks that gcc 4.9.2 doesn't support `-Wshift-overflow` and `-Wunused-const-variable`. It does support `-Warray-bounds` though. May be syntax for that flag got updated in latest `gcc`?

rockdaboot:
No problem, these are explicitly enabled warnings in `m4/wget_manywarnings.m4`.
Developers are expected to use the latest development tools :-)

Either remove .manywarnings, edit m4/wget_manywarnings.m4, or upgrade your totally outdated compiler ;-)

darnir:
@rockdaboot Didn't the `manywarnings` module check if the current compiler actually supports the flags? I remember, the Gnulib module we were using used to.

rockdaboot:
No, it uses `gcc -Q -help=warning,C` which is *much* faster than starting gcc for each possible warning. BUT... for these special flags, we add them hard-coded. Basically, we should only add them if gcc >= 6, if someone likes to work on that change (should be very simple).

rootkea:
So I removed `-Wshift-overflow=2` and `-Wunused-const-variable=2`from wget_manywarnings.m4. Also had to change `-Warray-bounds=2` to `-Warray-bounds`. Now `make` gets executed successfully. But now I see warning along the line `ISO C90 forbids mixed declarations and code`. Aren't we using `C99`?
I checked `./configure` output and it says:

    Compiler:           gcc -std=gnu99 -std=gnu99

So why C90 warnings?

rockdaboot:
Not sure. gcc 4.9 is C89 by default. Do a 'make clean' and a 'make V=1' to see the command line for compilation. Maybe that brings some light into it.

rockdaboot:
Just pushed a patch to check the gcc version... upstream m4/manywarnings.m4 should work for you now.

rootkea:
> Do a 'make clean' and a 'make V=1' to see the command line for compilation.

I see `-std=gnu99` on command line. Not sure why the `C90` warning though.
Should we explicitly add `-std=c99`?

rockdaboot:
No, that's not it. I can reproduce it here with a gcc 4.9.4.
It's a bug in gcc 4.9, the warning is enabled by -Wdeclaration-after-statement, even if set to C99.
Just ignore it for now, I see if there is an easy solution.

rockdaboot:
Pushed a fix to master now

rockdaboot:
@dwst What is the status with this issue right now ? Let me know if you need help.

dstw:
Sorry, I a little busy with the GSoC proposal. But, I have just commit the third version of this patch. It now reach 100% coverage.

dstw:
> BTW, have a look why Travis CI fails and try to fix it or ask if you are lost.

Still have no idea about Travis CI fails. I will start investigating. Is it problem with the code or I miss a Travis CI conf?

darnir:
We can't say that without looking at the logs from Travis. I'm currently traveling and unable to do so. But most probably it is an issue with to your commit since it breaks only for that. 

On 31 March 2017 10:00:15 AM GMT+02:00, Didik Setiawan <notifications@github.com> wrote:
>> BTW, have a look why Travis CI fails and try to fix it or ask if you
>are lost.
>
>Still have no idea about Travis CI fails. I will start investigating.
>Is it problem with the code or I miss a Travis CI conf?
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/rockdaboot/wget2/pull/155#issuecomment-290644261

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

dstw:
Now after I've done the last commit, Travis CI say that all checks have passed. It seems that indeed there is problems with my previous commit. Did I still need to check against it?

rockdaboot:
> Did I still need to check against it?

No. Next step would be to manually review your code. I'll find some time within the next hours.

rockdaboot:
Looks very good, just a few comments:
- you don't need 'struct test_entry', just put two arrays into 'struct test_data'.
- please add test data without 'sitemap'
- please add test data with 2 'sitemap' entries
- you have one test with two 'Disallow' entries, but you check just for one (the first)
- for completeness, you should also test if the number of parsed paths / sitemaps is equal to the number of test entries. Right now, if you expect 3 paths but only one is in robots->paths, you won't error

dstw:
> - you don't need 'struct test_entry', just put two arrays into 'struct test_data'.
> - please add test data without 'sitemap'
> - please add test data with 2 'sitemap' entries
> - you have one test with two 'Disallow' entries, but you check just for one (the first)
> - for completeness, you should also test if the number of parsed paths / sitemaps is equal to the number of test entries. Right now, if you expect 3 paths but only one is in robots->paths, you won't error

Done. Please review.

rootkea:
@dstw No you didn't.
Problem lies in `libwget/robots.c` Line 111
`for (p = data; !isspace(*p); p++);`

We need to change it to `for (p = data; *p && !isspace(*p); p++);`.
Can you spot why?

BTW I'm having similar feeling for line 103. Don't have time to go through the code in detail but maybe @rockdaboot can confirm.

dstw:
> Problem lies in libwget/robots.c Line 111
> for (p = data; !isspace(*p); p++);

How could you figure out this line is being source of the problem?

> Can you spot why?

mmm, let me guess, null problem? cmiiw

rootkea:
Travis log pretty much gave it away. See https://travis-ci.org/rockdaboot/wget2/jobs/217785106
More specifically:  
`#0 0x2b4894aba1ca in wget_robots_parse /home/travis/build/rockdaboot/wget2/libwget/robots.c:111:9`

And while traversing a string if our terminating condition is an (non)occurrence of a particular character other than `\0` then it's a standard code smell. That being said I don't know much about robots protocol so have no idea whether a `space` character is guaranteed at the end of string but which to me seemed illogical.

> null problem?

More specifically it's an illegal memory access.

rockdaboot:
Nice finding and a good lesson !
As @rootkea pointed out, Line 103 has the same problem.

@dwst To finish this PR:
- Add a test case that triggers the bug in line 103
- Merge your commits into one
- Add a fix for line 103 and 111, as @rootkea already proposed

Than your branch should have two commits and the PR is ready to be accepted.

rockdaboot:
Good work !
Rebased, merged ff-only and pushed manually.

PR #160, opened 2017-03-29
Fix path of valgrind-suppressions for separate build tree

akashrawal:
When build tree is different from source tree, libtest.la is not able to find suppressions file because it is looking in the build tree where the file is not there.


akashrawal:
Whoa, cannot reopen pull request after a rebase! Opening another one.

PR #161, opened 2017-03-29
Fix path of valgrind-suppressions for separate build tree

akashrawal:
When build tree is different from source tree, libtest.la is not able to find suppressions file because it is looking in the build tree where the file is not there.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -43,7 +43,8 @@ test_cookies_http_state_LDADD = ../src/log.o ../src/options.o libtest.la\
 noinst_LTLIBRARIES = libtest.la
 libtest_la_SOURCES = libtest.c
 libtest_la_CPPFLAGS = -I$(srcdir) -I$(top_srcdir)/include/wget -I$(top_builddir)/lib -I$(top_srcdir)/lib $(CFLAG_VISIBILITY) -DBUILDING_LIBWGET \
- -DWGETVER_FILE=\"$(top_builddir)/include/wget/wgetver.h\"
+ -DWGETVER_FILE=\"$(top_builddir)/include/wget/wgetver.h\" \
+ -DVGSUPPFILE=\"$(abs_srcdir)/valgrind-suppressions\"
 libtest_la_LIBADD = ../libwget/libwget.la\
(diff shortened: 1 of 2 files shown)


rockdaboot:
Just a question, why do you use $abs_srcdir and not $srcdir ?
We have that already as SRCDIR defined and available for the C sources.

rockdaboot:
Please amend the commit message to GNU standard. See `git log` for examples.

akashrawal:
I had tried that first, but found that tests are running from a subdirectory under $(builddir)/tests, This together with $(srcdir) being relative, is causing resultant path to be inaccurate.

rockdaboot:
We addressed that issue in other .c files in tests/, see e.g. libtest.c:
```
$ grep SRCDIR *.c
libtest.c:      // SRCDIR is the (relative) path to the tests dir. Since we chdir()'ed into a subdirectory, we need "../"
libtest.c:      wget_ssl_set_config_string(WGET_SSL_CA_FILE, "../" SRCDIR "/certs/x509-ca-cert.pem");
```

WDYT ?

akashrawal:
[Documentation](https://www.gnu.org/software/autoconf/manual/autoconf.html#Preset-Output-Variables) does not say anything about $(srcdir) being absolute or relative, so I assumed it could be either.

rockdaboot:
Since there is $(abs_srcdir) I always assumed $(srcdir) would be relative (many automake tutorials state this also). But you might be right. Then we better have to fix the use of SRCDIR, as in my above grep.

Thanks !

akashrawal:
I have produced a case where $(srcdir) is absolute. That's when I called configure with its absolute path. 

Doing that fails 'test-i-https'. 

    No CAs were found in '..//mnt/data/sw/contrib/wget/clone/wget2/tests/certs/x509-ca-cert.pem'

[test-suite.log.txt](https://github.com/rockdaboot/wget2/files/880189/test-suite.log.txt)

rockdaboot:
Oh, that is cool. And we know how to fix it now :-)

rockdaboot:
Just FYI, fixed in master now.

Issue #158, opened 2017-03-29
Fix test-bad-chunk


rockdaboot:
The sanitizer complains when switched on (`./configure --enable-fsanitize-asan --enable-fsanitize-ubsan`). The test already does a integer overflow check in runtime, so this is somewhat expected behavior. But it happens in at a line that should be protected by out check... so something is wrong there (or with the check).

From tests/test-bad-chunk.log:

```
29.121609.886 method 1 25 0:
29.121609.886 chunk size is 9223372036854775807
http.c:2480:19: runtime error: signed integer overflow: -9223262979044734957 - 109057810040857 cannot be represented in type 'long int'
Unexpected error code 1, expected 0 []
```

akashrawal:
The value of _p_ is incredibly large. In my machine MSB of _p_ is 1! Not yet sure where it is coming from, but subtracting two pointers results in signed integer type _ptrdiff_t_ which cannot store that result.
The error goes away by casting operands to _uintptr_t_ but I am not sure about portability of casting pointers to integers or availability of uintptr_t on all platforms, so no pull request.
https://github.com/akashrawal/wget2/commit/9144999388f5479b8635849816e5e2f0b1e135ec

rockdaboot:
Thanks for looking at it, any information is valuable at the moment. I'll have a closer look next week.

akashrawal:
Okay I see that gnulib is handling intptr_t and uintptr_t, so I guess we can assume casting to uintptr_t will not lose information in any system. Please correct me if I am wrong.

rockdaboot:
Both should be from `stdint.h`. That is covered by gnulib (module `stdint`).
The gnulib module list is here: https://www.gnu.org/software/gnulib/MODULES.html
I look it up pretty often...

akashrawal:
I see pointer type of `p` confused me there. I am concluding that pointer type of `p` from `libwget/http.c:2469` is just for comparison with other pointers. Making a pull request in a while.

rockdaboot:
Thanks for working on this !

PR #163, opened 2017-04-03
* libwget/http.c: Avoid signed integer overflow due to intermediate ptrdiff_t by casting operands to uintptr_t

akashrawal:
When two pointers are subtracted, the resultant is a signed integer `ptrdiff_t`  which might not be able to hold the resultant value. That is what is happening in bug #158.

This fixes bug #158.


rockdaboot:
Thanks !

PR #164, opened 2017-04-04
Use https:// instead of git://

rootkea:
Since most proxy/fire-walled environments block remote port 9418 which is used by `git://` protocol, use `https://` instead. This supports all the users including non-proxy environments.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "gnulib"]
 	path = gnulib
-	url = git://git.savannah.gnu.org/gnulib.git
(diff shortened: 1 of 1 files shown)


rootkea:
Or should we discard this as user end problem?

rockdaboot:
☺This file is private to you. Not part if the repository. Change it as you like.

rockdaboot:
Ups, too tired... Didn't have a coffee yet. Thought it was .git/config 😴

rockdaboot:
Very good catch !

PR #174, opened 2017-04-09
* configure.ac: Check for Valgrind and Sanitizers being mutually exclusive

### LINKED ISSUE #165: `./configure --enable-fsanitize-msan --enable-valgrind-tests && make check` completely freezes the system
`./configure --enable-fsanitize-msan --enable-valgrind-tests && make check` dooms the system. I didn't test but I think `./configure --enable-fsanitize-asan --enable-valgrind-tests` does the same.

We should:

1. Fail `./configure` when we see `asan/msan` and `valgrind`, spitting out the message of they being mutually exclusive instead of allowing it and then making the developer crying out blaming valgrind ;)
2. Most importantly *mention* this in `./configure --help`
I'm amazed `./configure --help` says nothing about even `asan` and `msan` being mutually exclusive. It's only when dev tries the actual command they find out this *after* the `./configure` fails!

rootkea:
* configure.ac: Check for valgrind and Sanitizers being mutually exclusive
* configure.ac: Make ./configure --help document above check
* configure.ac: Make ./configure --help document the existing check of asan and msan

This fixes #165

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -190,15 +190,15 @@ AC_ARG_ENABLE([fsanitize-ubsan],
   [gl_cc_sanitize_ubsan=yes], [gl_cc_sanitize_ubsan=no])
 
 AC_ARG_ENABLE([fsanitize-asan],
-  [AS_HELP_STRING([--enable-fsanitize-asan], [Turn on Address Sanitizer (for developers)])],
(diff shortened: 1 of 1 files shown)


rockdaboot:
Thanks, merging.

PR #175, opened 2017-04-09
* configure.ac: Check for msan and tsan being mutually exclusive

### LINKED ISSUE #173: Combining Memory and Thread sanitizers fails almost all of the tests
Since currently `./configure` can't recognize compiler being supporting `-fsanitize=thread`( #171) I used following command:

    $ CFLAGS="-g -fsanitize=thread -fPIC -pie" ./configure --enable-fsanitize-msan
This fails 29 of 31 tests. Here are few logs[0][1]

I think this is happening because `tsan` and `msan` aren't compatible and should be used in a mutually exclusive manner exactly like `asan` and `msan`.

If that's the case indeed  then we need to have a check for `tsan` and `msan` in `./configure` exactly like the existing check for `asan` and `msan`

[0] test-metalink.log: https://pastebin.com/zY5qKsU8
[1] test-parse-css.log: https://pastebin.com/mQXVYzfM

rootkea:
This fixes #173

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -210,6 +210,14 @@ if test "$gl_cc_sanitize_asan" = yes; then
   fi
 fi
 
+
+if test "$gl_cc_sanitize_msan" = yes; then
+  if test "$gl_cc_sanitize_tsan" = yes; then
+    AC_MSG_ERROR([Memory Sanitizer and Thread Sanitizer are mutually exclusive])
+  fi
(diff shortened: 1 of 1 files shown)


rootkea:
I'll add help string once previous PR gets merged to avoid merge conflicts.

rockdaboot:
Thanks !

PR #179, opened 2017-04-11
Prevent wget2rc from being read during tests

darnir:
* src/options.c: Treat parse_filename{,s} as string vectors and clear
them with a no-* command line option
* tests/libtest.c: Pass --no-config to all the tests by default


rockdaboot:
I just restarted the job .1. `make check` was hanging...

rockdaboot:
Thanks, pushed manually.

Issue #170, opened 2017-04-05
cut-dirs is broken


5hd:
As I understand it, currently the "cut-dirs" setting is broken.
Issues I encountered:
+ --cut-dirs=1 together with --no-host-directories will fail to save any file
+ leading / should be removed

See my commit here: https://github.com/MichaelHeerklotz/wget2/commit/fb665a12d3d1b8a1d4154847ed5f1350741fd8bf

rockdaboot:
(Sorry if you get this twice (also put as comment to your commit))

Thanks for having a look at it.
Please create a pull request (PR) - then your changes are automatically checked by our Continuous Integration (Travis CI). If that passes, I'll do a manual review.

If you have time and joy, please add a test case to reproduce a/the problem (see tests/test.*.c as examples).

MichaelHeerklotz:
Okay, I will give it a try in a few days, if I find the time.

rockdaboot:
Fixed by #185

PR #185, opened 2017-04-11
fixed and added test for cut-dirs

MichaelHeerklotz:
see https://github.com/rockdaboot/wget2/issues/170


rockdaboot:
Good work ! And thanks for the new test.

I slightly amended the code (cosmetics) and rewrote the commit message. We need it in that special GNU style to be able to auto-generate the ChangeLog file for distribution tarballs. Please try to format the commit message appropriate in the future (see `git log` for examples).

Pushed to master.

PR #186, opened 2017-04-12
added setting "--cut-get-vars" to cut http get vars from file names

MichaelHeerklotz:


darnir:
Thanks for working on this! This is definitely a good feature to have. However, could you please also add a test case to ensure coverage?

Even better, if you could split it into a separate function and have a unit test for just the function.

rockdaboot:
What @darnir says 👍 

Just a question for my understanding... is this intentionally for retrieving only ?
E.g. `http://domain?getpng&id=5` would retrieve the `index.html` instead of the expected PNG !?

Maybe you could explain your use case or give an example !?
The best is your put this example/explanation to `docs/wget2_manual.md` ☺️

tests/test-cut-get-vars.c
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright(c) 2017 Michael Heerklotz
```
rockdaboot: 2017 Should be for the FSF only (remove the first (C) line and set the year to 2017 for FSF).

Explanation:
I transferred my copyrights in 2015 to the FSF, so there are files with (C) Tim Rühsen + (C) FSF - the copyrigth year for the FSF should be >= 2015 and my copyright year should be <= 2015.

src/wget.c
```diff
@@ -340,6 +340,14 @@ const char * G_GNUC_WGET_NONNULL_ALL get_local_filename(wget_iri_t *iri)
 	} else {
 		fname = wget_iri_get_filename(iri, &buf, config.local_encoding);
 	}
+	
+	if(config.cut_file_get_vars) {
```
rockdaboot: Please add a space between if and (. At other places as well.

src/wget.c
```diff
@@ -340,6 +340,14 @@ const char * G_GNUC_WGET_NONNULL_ALL get_local_filename(wget_iri_t *iri)
...
+		char *p = strchr(fname, '?');
+		
+		if(p) {
```
rockdaboot: We normally do not use curly brackets around a single line of code.

src/wget.c
rockdaboot: If config.cut_file_get_vars is set, you can skip the call wget_iri_get_query_as_filename() a few lines above (fname = buf.data instead).
Also you could call wget_iri_get_path() instead of calling wget_iri_get_filename().
That just avoids appending the query to the filename.

MichaelHeerklotz:
Thank you for the feedback, I will adjust the code soon.

MichaelHeerklotz:
I have updated the code.

The reason for this setting is: When crawling pages that have css files and sprite images with "export tags", these should be stripped.
For example "/main.css?v=1" or "/sprite.png?v=7".
I have a use case for wget2 where many pages are exported from a CMS to generate more or less "static" pages. For these pages to work files like "main.css?v=x" should be saved without the get vars.

docs/wget2_manual.md
```diff
@@ -673,6 +673,19 @@ Go to background immediately after startup. If no output file is specified via t
...
+  For example "main.css?v=123" will be changed to "main.css".
+  Be aware that this may have unintended side effects, for example "image.php?name=sun" will be changed 
+  to "image.php". The cutting happens before adding the URL to the download queue.
+
+* --cut-file-get-vars
+
+  Remove HTTP GET Variables from filenames.
+  For example "main.css?v=123" will be changed to "main.css".
+  Be aware that this may have unintended side effects, for example "image.php?name=sun" will be changed 
+  to "image.php". The cutting happens when saving the file, after downloading.
```
darnir: A small segment here about how it works along with `--trust-server-names` and `--content-disposition` would be nice. Especially a hint that using `--content-disposition` _may_ prevent the side effects that are shown above

MichaelHeerklotz:
Thank you for the feedback, I have updated the manual.

rockdaboot:
Pushed manually, thanks for the good work !

PR #187, opened 2017-04-12
Added regex filter settings and new filter-urls setting

MichaelHeerklotz:
Added regex filters (--accept-regex, --reject-regex, --regex-type) like in wget1, doc was already there, but the code for it was missing.

Added the new setting "--filter-urls" which will apply the filter on the URL before adding the URL to the job queue.

I also updated the doc for "--accept-regex" and "--reject-regex" to clarify that these are applied on the file name like "--accept and --reject" by default (if --filter-urls is off).

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -594,6 +594,22 @@ static int G_GNUC_WGET_PURE G_GNUC_WGET_NONNULL((1)) parse_cert_type(option_t op
 	return 0;
 }
 
+static int G_GNUC_WGET_PURE G_GNUC_WGET_NONNULL((1)) parse_regex_type(option_t opt, const char *val)
+{
+	if (!val || !wget_strcasecmp_ascii(val, "posix"))
+		*((char *)opt->var) = WGET_REGEX_TYPE_POSIX;
+		
+#if defined(WITH_LIBPCRE2) || defined(WITH_LIBPCRE)
+	else if (!wget_strcasecmp_ascii(val, "pcre"))
+		*((char *)opt->var) = WGET_REGEX_TYPE_PCRE;
+#endif
+
+	else
+		error_printf_exit("Unsupported regex type '%s'\n", val);
+
+	return 0;
+}
+
 static int G_GNUC_WGET_PURE G_GNUC_WGET_NONNULL((1)) parse_progress_type(option_t opt, const char *val)
 {
 	if (!val || !*val || !wget_strcasecmp_ascii(val, "none"))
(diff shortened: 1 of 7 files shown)


src/options.c
```diff
@@ -823,7 +833,9 @@ static const struct optionw options[] = {
 	{ "read-timeout", &config.read_timeout, parse_timeout, 1, 0 },
 	{ "recursive", &config.recursive, parse_bool, 0, 'r' },
 	{ "referer", &config.referer, parse_string, 1, 0 },
+	{ "regex-type", &config.regex_type, parse_string, 1, 0 },
```
darnir: Preferably define a default value for `config.regex_type`.

src/wget.c
```diff
@@ -449,6 +460,99 @@ static int in_host_pattern_list(const wget_vector_t *v, const char *hostname)
...
+  	pcre_free(re);
+
+	return result;
+}
+#endif
+
+static int regex_match(const char *string, const char *pattern)
+{
+#if defined(WITH_LIBPCRE2) || defined(WITH_LIBPCRE)
+	if(config.regex_type && !wget_strcmp(config.regex_type, "pcre"))
```
rockdaboot: if( :smiley:

MichaelHeerklotz: upps... ;)

rockdaboot:
@MichaelHeerklotz This is so cool to see !

As @darnir says, a check for valid regex-type is user friendlier then silently falling back to posix.
I guess you need an own parser function like `parse_prefer_family()`, e.g. a parse_regex_type().
Together with some defines, e.g. WGET_REGEX_TYPE_POSIX, WGET_REGEX_TYPE_PCRE.

Can't wait to merge it !

MichaelHeerklotz:
Thank you for the nice feedback! I have updated the regex_type setting.

MichaelHeerklotz:
The new option "--filter-urls" currently will filter URL **and** filename, which gives me some trouble.
I think, this should be improved.

The question is, do we want to behave like old wget, or do we want to improve the situation? 

In my use case, --cut-dirs=1 is used together with --regex-accept, where the regex matches the first path in the URL. This works with wget1, but fails with wget2 (with and without this patch).

An idea would be to have separate --accept(-regex) and --reject(-regex) settings for URLs and file names.
For example:
Match file names:  --accept(-regex), --reject(-regex) 
Match URLs: --accept-url(-regex), -reject-url(-regex)

Or we could just change the behavior of --filter-urls in this patch, to only check the URL when it is enabled.
The downside would be that one has to decide whether to filter file names or URLs.

What do you think?

rockdaboot:
> Match file names: --accept(-regex), --reject(-regex)
Match URLs: --accept-url(-regex), -reject-url(-regex)

I find this is basically a good idea. During time I saw several requests for such fine grained control for wget1.x (there might also be open issues on the Savannah bug tracker). Filtering for files also gives us better control for Content-Disposition with filename=.

Still we have to stay backwards compatible as good as we can. So maybe
Match URLs: --accept-url(-regex), -reject-url(-regex)
Match files: --accept-file(-regex), -reject-file(-regex)
While the old --accept(-regex) and --reject(-regex) work as before !?

That doesn't solve the question what comes first, --cut-dirs and than file-filter or the other way round.

While writing I remember that some people requested something like
--accept-mime-type / --reject-mime-type

Also, we should think about having the filter/RE type before each rule and a optional +/-, e.g.
--filter-mime-type=NONE,+[PCRE|POSIX]:^audio/*wav*
or
--filter-mime-type=ALL,-[PCRE|POSIX]:^audio/*wav*

I have to open another issue ;-)

rockdaboot:
Worked on at upstream (Gitlab).

PR #188, opened 2017-04-13
Added a more detailed logging for decompress and HTTP errors

MichaelHeerklotz:
* libwget/http.c: Now logging host and resource on decompress error
* src/wget.c: More detailed logging for HTTP response code

### DIFF
--- a/src/wget.c
+++ b/src/wget.c
@@ -1218,7 +1218,10 @@ static int process_response_header(wget_http_response_t *resp)
 	DOWNLOADER *downloader = job->downloader;
 	wget_iri_t *iri = job->iri;
 
-	print_status(downloader, "HTTP response %d %s\n", resp->code, resp->reason);
+	if (resp->code < 400 || resp->code > 599)
+		print_status(downloader, "HTTP response %d %s [%s]\n", resp->code, resp->reason, iri->uri);
+	else
+		print_status(downloader, "HTTP ERROR response %d %s [%s]\n", resp->code, resp->reason, iri->uri);
 
 	// Wget1.x compatibility
 	if (resp->code/100 == 4) {
--- a/libwget/http.c
+++ b/libwget/http.c
@@ -90,6 +90,14 @@ static wget_vector_t
 	*https_proxies,
 	*no_proxies;
 
+static void call_wget_decompress(wget_http_response_t *resp, wget_decompressor_t *dc, char *src, size_t srclen)
+{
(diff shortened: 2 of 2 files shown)


src/wget.c
```diff
@@ -1218,7 +1218,10 @@ static int process_response_header(wget_http_response_t *resp)
 	DOWNLOADER *downloader = job->downloader;
 	wget_iri_t *iri = job->iri;
 
-	print_status(downloader, "HTTP response %d %s\n", resp->code, resp->reason);
+	if(resp->code == 200 || resp->code == 301 || resp->code == 302 || resp->code == 304)
```
darnir: I don't think this is a very good idea. Handling individual response codes will soon get too cumbersome. Currently, this treats a 206 Partial Content as an `ERROR` as well. 

Pretty much only the 4xx and 5xx series of responses are errors (Client and Server, respectively). If you'd really like this in the output, please match on the series instead of individual response code values

rockdaboot:
I know it's sometimes tedious to split things. But please try to separate independent things into different commits, and preferable also into several PRs. That makes the review process much easier for us and for you (if we request changes). It also allows faster merging in the end.

MichaelHeerklotz:
I have updated the code :)

rockdaboot:
@MichaelHeerklotz I just realized, that we have a 'context' that we give to wget_decompress_open(). This context is already has all the information needed (wget_http_response_t *). All we need (and this seems more elegant to me) would be an error handler function. It takes the context as argument, so it can print out any (custom) error message. We could set this handler via  wget_decompress_open() or (maybe better) with a new function wget_decompress_set_error_handler(). This would be an enrichment to the decompress API which might come anyways in the future.
I am sorry that I didn't think about it earlier.

MichaelHeerklotz:
@rockdaboot This makes sense. Should we remove the error messages in decompressor.c (like "Failed to uncompress gzip stream"), or maybe still print them, when no error handler callback was registered?

rockdaboot:
> maybe still print them, when no error handler callback was registered

I think that is perfect for now. If someone requests absolutely silence later, it needs another function (...set_flags() or something).

darnir:
Is this PR ready for merging?

rockdaboot:
@MichaelHeerklotz We moved to Gitlab, please consider opening an account there). You can work on your open PRs here or there, but please open new issues/PRs on Gitlab only.

PR #189, opened 2017-04-13
Fixed crash in convert-links code introduced with html css parsing.

MichaelHeerklotz:
* include/wget/wget.h: Removed unused variable.
* libwget/html_url.c: Updated code to always return pointer inside "html" variable.
* test/test-k.c: Updated test to cover the bug.

Sorry, my HTML CSS patch introduced a crash bug, because the URL pointer is converted back to a offset inside the html variable when the "-k" option is used. This patch should take care of this problem.
I also updated the test to cover this issue.

### DIFF
--- a/include/wget/wget.h
+++ b/include/wget/wget.h
@@ -1233,8 +1233,6 @@ typedef struct {
 	wget_string_t
 		url;
 	char
-		must_free_url;
-	char
 		attr[16];
 	char
 		dir[16];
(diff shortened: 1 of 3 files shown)


rockdaboot:
Push locally, many thanks for
* making the code even better (removed malloc(), condensed cleanup)
* finding and fixing the bug
* adding a test case
👍

PR #190, opened 2017-04-13
WIP: Replace INET_FAMILY #defines with enums

darnir:
## Please do not merge this PR

This is a WIP for an idea I had and want to see what the rest of you think about it. Currently, for some options, we use a set of `#define` macros and pass around the values as an `int`. It would be cleaner to group them correctly as an enumeration and pass it around as such. This also makes the code more readable since it is very obvious what the data type is.

If everyone agrees on this format, I'll spend some more time to migrate the other options as well.

### DIFF
--- a/include/wget/wget.h
+++ b/include/wget/wget.h
@@ -1319,9 +1319,11 @@ WGETAPI void
  * TCP network routines
  */
 
-#define WGET_NET_FAMILY_ANY  0
-#define WGET_NET_FAMILY_IPV4 1
-#define WGET_NET_FAMILY_IPV6 2
+typedef enum {
+    WGET_NET_FAMILY_ANY,
+    WGET_NET_FAMILY_IPV4,
+    WGET_NET_FAMILY_IPV6
+} WGET_INET_FAMILY;
 
 #define WGET_PROTOCOL_HTTP_1_1  0
 #define WGET_PROTOCOL_HTTP_2_0  1
--- a/libwget/net.c
+++ b/libwget/net.c
@@ -307,7 +307,7 @@ struct addrinfo *wget_tcp_resolve(wget_tcp_t *tcp, const char *host, const char 
 	return addrinfo;
 }
 
-static int G_GNUC_WGET_CONST _value_to_family(int value)
+static int G_GNUC_WGET_CONST _value_to_family(WGET_INET_FAMILY value)
 {
 	switch (value) {
 	case WGET_NET_FAMILY_IPV4:
@@ -319,7 +319,7 @@ static int G_GNUC_WGET_CONST _value_to_family(int value)
 	}
 }
 
-static int G_GNUC_WGET_CONST _family_to_value(int family)
+static WGET_INET_FAMILY G_GNUC_WGET_CONST _family_to_value(int family)
 {
 	switch (family) {
 	case AF_INET:
--- a/libwget/net.h
+++ b/libwget/net.h
@@ -50,7 +50,6 @@ struct wget_tcp_st {
 		connect_timeout,
 		timeout, // read and write timeouts are the same
 		family,
-		preferred_family,
 		protocol; // WGET_PROTOCOL_HTTP1_1, WGET_PROTOCOL_HTTP2_0
 	unsigned char
 		ssl : 1,
@@ -61,6 +60,8 @@ struct wget_tcp_st {
 		tls_false_start : 1,
 		tcp_fastopen : 1, // do we use TCP_FASTOPEN or not
 		first_send : 1; // TCP_FASTOPEN's first packet is sent different
+	WGET_INET_FAMILY
+		preferred_family;
 };
 
 #endif /* _LIBWGET_NET_H */
--- a/src/wget_options.h
+++ b/src/wget_options.h
@@ -119,7 +119,6 @@ struct config {
 		waitretry,
 		restrict_file_names,
 		level,
-		preferred_family,
 		cut_directories,
 		connect_timeout, // ms
 		dns_timeout, // ms
@@ -190,6 +189,8 @@ struct config {
 		metalink,
 		cut_url_get_vars,
 		cut_file_get_vars;
+	WGET_INET_FAMILY
+		preferred_family;
 };
 
 extern struct config


MichaelHeerklotz:
I like the idea.  It is easier to read, and this way it is harder to pass an invalid type/value.
Some projects I have worked on in the past had the rule to give every enum value a specific int value.
For example
```
typedef enum {
    WGET_NET_FAMILY_ANY = 0,
    WGET_NET_FAMILY_IPV4 = 1,
    WGET_NET_FAMILY_IPV6 = 2
} WGET_INET_FAMILY;
``` (should be a part of the C99 standard)
This way the values passed around stay the same, even if the order is changed or a value is deleted, much like when using defines.

**Comment by darnir** [comment]
> This way the values passed around stay the same, even if the order is changed or a value is deleted, much like when using defines.

If we depend on the values when using an enum, we're doing it wrong. One should always match on the enum elements not the underlying value. The whole point of switching out of defines is to prevent that.

**Comment by MichaelHeerklotz** [comment]
Yes, the int values should never be used directly.
But in the background the compiler will always generate an int value which will be passed around, even if we do not explicitly set one. This does not really matter, but when there are external interfaces involved,  bindings to other languages exist, or there is a header mismatch, then it might be a good thing.

**Comment by rockdaboot** [comment]
I don't have any problems with #defines, but had several with enums in the past. But these are no blockers. But I can follow the 'readability' argument. And I hope for better/stricter enum support in future versions of C (e.g. strict enum type checking and removing the 'global' attitude).

I experienced some really bad 'we move #defines to enums' in the past, e.g. with GnuTLS. We had code that used `#ifdef new_feature` - when GnuTLS then moved that #define to an enum, guess what happened !? We silently used old/obsolete code until one day I stumbled upon it.
Conclusion: Better use enums in public headers now than moving later (after the first release).

BTW, the integer type of an enum is not specified in C99 and varies with compiler/version and likely with -O flags (allowed is char and any signed or unsigned integer type). That is bad when using in structures - it might blow up the memory footprint (at least gcc gas __attribute__ ((__packed__)) to avoid that). So enums take away a bit of control and might also introduce random warnings when assigning / comparing with abitrary integer values. Using enums as array index is also problematic, you might get warnings when the enum is a char type. You don't have these problem with #defines - I just want to mention it, please think about it.

IMO: When want to have enums, we should use typedefs and do it ASAP before the first release.
@darnir If you go ahaead, just create a new branch in thsi repo so anybody can drop in for helping.

**Comment by darnir** [comment]
@rockdaboot: The enums I'm looking to change are all from the `options.c` file, so they will most probably not be used in a struct.

About comparing with integers, my point was that we should not be using these values and comparing them to an integer / using them to index into an array.

I'll start working on changing as many as I can and will keep pushing to this PR. Once we've agreed on it, I'll rebase and make a clean commit.

BTW, do you agree with the idea from @MichaelHeerklotz? To use explicit integer assignments? That makes sure that the enum is always an integer doesn't it? I'm not sure if the spec allows for that or if it is a GCC / LLVM extension

**Comment by darnir** [comment]
I see a couple of places where the WGET_NET_FAMILY_* variables are used in the codebase assuming that they are integers. In some places, like, `wget_global_get_int()`, in `libwget/init.c`, explicitly rely on these values being ints. 
How do you envision this being replaced to use an enum?

**Comment by rockdaboot** [comment]
To use explicit integer assignments?

Yes, that's preferable.

> That makes sure that the enum is always an integer doesn't it?

No it doesn't. The compiler is free to chose a fitting integer type or char, depending on the values and/or optimization level. -Os and -O2 might result in a different type.

> `int wget_global_get_int(int key)`

`key` is primarily what we are taking about. I see no problem using an enum type. The return value is an integer number - we possibly have to cast it, e.g. to boolean or the corresponding enum type. That depends on key. The alternative is having many of these functions... we should not have too many functions doing basically the same.

**Comment by rockdaboot** [comment]
Warnings resulting from different integer type from (clang, FreeBSD 11):
```
hashfile.c:474:16: warning: comparison of unsigned enum expression >= 0 is always true [-Wtautological-compare]
        if (algorithm >= 0 && algorithm < countof(_algorithm))
```

On gcc enums are signed, with clang they are unsigned. The enum is `wget_digest_algorithm_t`.

Removing the first expression would crash (or overwrite memory) on gcc if someone gives -1 to wget_hash_get_len(). This even might result in a severe security flaw.

Using -Wno-tautological-compare doesn't warn us on certain kinds of bad code.

The C standard gives us nothing to deal with this situation gracefully. Whatever we do feels like a hack. From a technical standpoint, an `int ` seems to be a better choice than an `enum`. But you threw in the `readability` argument which is also very important.

**Comment by darnir** [comment]
I've been looking into these differences as well. At this point, I think I'm willing to throw the "readability" argument in the bin and stick with `int`s. This is mostly because like you said, anything we do looks like a hack. Which on its own reduces the readability of the code. I think it is far more obvious like this than with enums after we implement all the hacks required to make sure the code is always correct.

**Comment by rockdaboot** [comment]
What if we cast the enum value to have a defined type, .e.g.
```
	if (algorithm >= 0 && algorithm < countof(_gcrypt_algorithm)) {
```

and we change it to:
```
	if ((unsigned) algorithm < countof(_gcrypt_algorithm)) {
```

**Comment by MichaelHeerklotz** [comment]
I would remove the "_gcrypt_algorithm" array and use a function with a switch case instead.

**Comment by rockdaboot** [comment]
> I would remove the "_gcrypt_algorithm" array and use a function with a switch case instead.

IMO, that is inelegant code bloating and it doesn't scale (what do you do if you have 100 different values ?).  It also increases execution time whenever the compiler can't or won't optimize your switch to an array access. That means in some cases the effort increases from O(1) to O(N).

If you do things like that in general, be prepared to have bug reports like "On obscure system X with even more obscure compiler Y, wget became somewhat slow... the old version from 2005 was so much faster". Fix that if you don't get any more information from the reporter ;-)

**Comment by MichaelHeerklotz** [comment]
I have nerver seen any compiler not generating a lookup-array for a switch case.
Neither Visual C, nor old gcc versions from like 1995 did it.

**Comment by rockdaboot** [comment]
> I have nerver seen any compiler not generating a lookup-array for a switch case.

So you did test with all option combinations and observed the assembler output !?

There may be ~50 C compilers out there, not counting the different versions.
At least I saw many C compilers not optimizing at all (or hardly even when requested). Many compilers have a switch to disable any kind of optimization... if they still optimize to arrays, it is a bug !

**Comment by MichaelHeerklotz** [comment]
Okay, but your code
`if ((unsigned) algorithm < countof(_gcrypt_algorithm)) {`
is also not that perfect, because if you forget the (unsigned)  when using it as an index a few lines below, the check is useless.
better store the casted value in a unsigned variable

**Comment by rockdaboot** [comment]
Just tested enums translations with gcc... there is no array code generated with `gcc switch.c -S`, with switch.c being:
```
#include <stdio.h>

typedef enum {
  my_a = 1,
  my_b = 2,
  my_c = 3,
  my_d = 4,
  my_e = 5,
} my_t;

int main(int argc, char **argv)
{
  my_t x = argv[1][0] - '0';

  switch (x)
    {
    case my_a:
      printf("1\n");
      break;
    case my_b:
      printf("2\n");
      break;
    case my_c:
      printf("3\n");
      break;
    case my_d:
      printf("4\n");
      break;
    case my_e:
      printf("5\n");
      break;
    default:
      printf("What the hell !\n");
    }

  return 0;
}
```

gcc (6.3.0 here) produces a jump table to the corresponding blocks of code, but not a direct array access. Look into the assembler code.

**Comment by rockdaboot** [comment]
> Okay, but your code
> > if ((unsigned) algorithm < countof(_gcrypt_algorithm)) {

> is also not that perfect, because if you forget the (unsigned) when using it as an index a few lines below, the check is useless.

**Comment by rockdaboot** [comment]
switch.c extended with the questionable check and direct array access:
```
#include <stdio.h>

typedef enum {
  my_a = 1,
  my_b = 2,
  my_c = 3,
  my_d = 4,
  my_e = 5,
} my_t;

static char *msg[] = {
  "1",
  "2",
  "3",
  "4",
  "5"
};

int main(int argc, char **argv)
{
  my_t x = argv[1][0] - '0';

  switch (x)
    {
    case my_a:
      printf("1\n");
      break;
    case my_b:
      printf("2\n");
      break;
    case my_c:
      printf("3\n");
      break;
    case my_d:
      printf("4\n");
      break;
    case my_e:
      printf("5\n");
      break;
    default:
      printf("What the hell !\n");
    }

  if ((unsigned) (x - 1) < sizeof(msg)/sizeof(msg[0]))
    printf("%s\n", msg[x - 1]);
  else
    printf("What the hell !\n");

  return 0;
}
```

The assembler code is **much** smaller:
```
        movl    -4(%rbp), %eax
        subl    $1, %eax
        movl    %eax, %eax
        leaq    0(,%rax,8), %rdx
        leaq    msg(%rip), %rax
        movq    (%rdx,%rax), %rax
        movq    %rax, %rdi
        call    puts@PLT
```

And you can test `switch` with any value. The array code is perfectly protected (try values \< 1 or \> 5).

rockdaboot:
Btw, `-O2` doesn't optimize the enum to array access either, but the direct array access becomes optimized even shorter.

MichaelHeerklotz:
a jump table should be pretty fast

MichaelHeerklotz:
yes, you are right, i was wrong with the assumption that the access would overflow

PR #191, opened 2017-04-17
Ignore all anchor links instead of only '#'

MichaelHeerklotz:
* examples/check_url_types.c: Modified _normalize_uri() to ignore all anchor links
* examples/print_css_urls2.c: Modified css_parse_uri() and _convert_links() to ignore all anchor links
* src/wget.c: Modified _normalize_uri() and _convert_links() to ignore all anchor links

I think we should ignore all anchor links, not just only '#', or am I missing something?

### DIFF
--- a/examples/check_url_types.c
+++ b/examples/check_url_types.c
@@ -70,7 +70,7 @@ static int _normalize_uri(wget_iri_t *base, wget_string_t *url, const char *enco
 	size_t urlpart_encoded_length;
 	int rc;
 
-	if (url->len == 0 || (url->len == 1 && *url->p == '#')) // ignore e.g. href='#'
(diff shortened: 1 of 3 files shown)


MichaelHeerklotz:
Now when I think about it, the HTML URL scanner should probably ignore anchor links, too. I'll add the code and add a test for it, when I find the time.

rockdaboot:
The APP should decide what To do...

rockdaboot:
Sorry, was just walking through the city... :-)

The URL scanner can't decide what the application is looking for.
Please remove the check from print_css_urls2.c, it is useless or even irritating there.
And then it's ready to be merged.

rockdaboot:
Thanks ! Slightly amend and merged manually.

PR #194, opened 2017-04-21
* src/options.c (print_version): Remove redundant preprocessor check

rootkea:
Still not sure why do we need to show `"ssl/gnutls"` when we are already showing `"https"`.


rockdaboot:
Basically for backward compatibility with wget1.x (some tools parse the output of --version).
That is why we also have -ntlm and -opie though we don't support both (yet).

PR #197, opened 2017-09-28
Terminate http request on "416 Requested Range Not Satisfiable" response status code.

jmoellers:
When a "416" Response Status Code is received, the request is terminated with an error message:
HTTP ERROR response 416 Requested Range Not Satisfiable [<URL>]
This sometimes happens when a server responds incorrectly to a "continue" request.
There is not enough information available (Range of the request) to properly handle this case. If it were, the case that the file has already been completely downloaded could be detected and an appropriate message could be generated.
I also took the liberty to port the status codes from the old wget source to libwget/http.h to make the code more readable.


rootkea:
@jmoellers Thank you for your work but could you please open a pull request (in gitlab-speak "merge request") at https://gitlab.com/gnuwget/wget2 ?

@rockdaboot May be we should amend "Clone of the GNU Wget2 repository for github collaboration" - github repo description, to reflect the same? (Not many people will read README.md thoroughly and thus will most probably miss "Development" link)

rockdaboot:
@rootkea Any suggestions ? I keep the Github repo close to the Gitlab repo - people often tend to first search at Github for a project and if they work on it, their work is based on pretty recent clone. So moving the work to Gitlab is very easy. And we 'catch' those contributors from here, just with a polite hint :-)

rootkea:
umm.. But we can save the inconvenience of moving their work to Gitlab. Moreover the MR itself will get benefit from peer review if opened at GitLab in the first place.

How about "This is a mirror of Gitlab repo. Hack wget2 at https://gitlab.com/gnuwget/wget2" or something along those lines? "Mirror of GitLb repo. Development happens at https://gitlab.com/gnuwget/wget2"

rockdaboot:
Thanks, changed the wording now.

PR #229, opened 2020-10-11
* .travis.sh: Output tail -1000 of test logs

rockdaboot:
Should fix termination of Travis-CI due to too many lines of output.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -38,9 +38,9 @@ for OPTS in "${CONFIGURE_OPTIONS[@]}"; do
 		make install -j3
 	fi
 	if make clean check -j3; then :; else
-		test -f fuzz/test-suite.log && cat fuzz/test-suite.log
-		test -f unit-tests/test-suite.log && cat unit-tests/test-suite.log
-		test -f tests/test-suite.log && cat tests/test-suite.log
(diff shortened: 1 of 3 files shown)


.travis.sh
```diff
@@ -38,9 +38,9 @@ for OPTS in "${CONFIGURE_OPTIONS[@]}"; do
 		make install -j3
 	fi
 	if make clean check -j3; then :; else
-		test -f fuzz/test-suite.log && cat fuzz/test-suite.log
-		test -f unit-tests/test-suite.log && cat unit-tests/test-suite.log
-		test -f tests/test-suite.log && cat tests/test-suite.log
+		test -f fuzz/test-suite.log && tail -1000 fuzz/test-suite.log
+		test -f unit-tests/test-suite.log && tail -1000 unit-tests/test-suite.log
+		test -f tests/test-suite.log && tail -1000 tests/test-suite.log
```
emilengler, line 43: `-1000` doesn't look specified somehow. Shouldn't you use `-n 1000` instead?

rockdaboot: Good point. I use an old, obsolete syntax here (still mentioned in the coreutils online manual).
Will change it to -n as soon as I get back to the keyboard :-)

PR #280, opened 2023-06-14
Win32 enhancements

mitchcapper:
Some changes I made to track down some bugs using MSVC.  Decently enhanced windows support as well.  If there is interest in any of these let me know, I can put PRs in for whichever individually and certainly open to any refactoring desired.   Many are Windows specific but several are not, and a few are decently broken fixes.  Id look at each commit rather than just the massive diff blob.

Some of these things may specifically not be wanted.  It is pretty clear wget2 made a deliberate choice to require a c99 compiler for dynamic arrays.  This most notably excludes MSVC.   I wanted to track down a use after free bug so went to do so.  If bringing MSVC support back is of interest the code could be improved (say MSVC only #ifdef gating around dynamic arrays or use alloca rather than malloc).   There are 13 files and probably 20 some uses that had to be changed but it was fairly minor ( https://github.com/rockdaboot/wget2/commit/f0adf4a3e6d0babd2791bc9aad7a2a485312a322 ).

There was also an attempt to remove some spurious or excessive error log messages.  

I also added this as part of a minor side project ( https://github.com/mitchcapper/WIN64LinuxBuild ) for automated native MSVC compiling of various utils.

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -14,7 +14,7 @@ noinst_PROGRAMS = \
 
 
 AM_CFLAGS = $(WERROR_CFLAGS) $(WARN_CFLAGS) -DMALLOC_RETURNS_NONNULL
-AM_CPPFLAGS = -I$(top_srcdir)/include/wget \
+AM_CPPFLAGS = -I$(top_srcdir)/lib -I$(top_srcdir)/include/wget \
  -DWGETVER_FILE=\"$(top_builddir)/include/wget/wgetver.h\" -DMALLOC_RETURNS_NONNULL
 AM_LDFLAGS = -no-install
 LDADD = ../libwget/libwget.la\
--- a/examples/batch_loader.c
+++ b/examples/batch_loader.c
@@ -20,7 +20,7 @@
  * Read URLs from stdin and download into results/domain/.
  *
  */
-
+#include "config.h"
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
--- a/libwget/console.c
+++ b/libwget/console.c
@@ -99,7 +99,23 @@ void wget_console_reset_fg_color(void)
 {
 	wget_console_set_fg_color(WGET_CONSOLE_COLOR_RESET);
 }
+#ifdef _WIN32
+static DWORD SetupConsoleHandle(BOOL is_input, HANDLE handle) {
+	DWORD mode = 0;
+	if (handle == INVALID_HANDLE_VALUE)
+		return mode;
+	if (!GetConsoleMode(handle, &mode))
+		return mode;
+	DWORD orig = mode;
+	if (is_input)
+		mode |= ENABLE_VIRTUAL_TERMINAL_INPUT;
+	else
+		mode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
 
+	SetConsoleMode(handle, mode);
+	return orig;
+}
+#endif
 /**
  * \return 0 on success, or -1 on error
  *
@@ -120,7 +136,8 @@ int wget_console_init(void)
 		if (GetFileType(g_stdout_hnd) != FILE_TYPE_CHAR) /* The console is redirected */
 			g_stdout_hnd = INVALID_HANDLE_VALUE;
 	}
-
+	SetupConsoleHandle(true, GetStdHandle(STD_INPUT_HANDLE));
+	SetupConsoleHandle(false, GetStdHandle(STD_OUTPUT_HANDLE));
 	win_init = 1;
 #endif
 
--- a/libwget/dns.c
+++ b/libwget/dns.c
@@ -19,7 +19,28 @@
  *
  * resolver routines
  */
-
+#ifdef __cplusplus
+    #define INITIALIZER(f) \
+        static void f(void); \
+        struct f##_t_ { f##_t_(void) { f(); } }; static f##_t_ f##_; \
+        static void f(void)
+#elif defined(_MSC_VER)
+    #pragma section(".CRT$XCU",read)
+    #define INITIALIZER2_(f,p) \
+        static void f(void); \
+        __declspec(allocate(".CRT$XCU")) void (*f##_)(void) = f; \
+        __pragma(comment(linker,"/include:" p #f "_")) \
+        static void f(void)
+    #ifdef _WIN64
+        #define INITIALIZER(f) INITIALIZER2_(f,"")
+    #else
+        #define INITIALIZER(f) INITIALIZER2_(f,"_")
+    #endif
+#else
+    #define INITIALIZER(f) \
+        static void f(void) __attribute__((constructor)); \
+        static void f(void)
+#endif
 #include <config.h>
 
 #include <sys/types.h>
@@ -68,20 +89,20 @@ static wget_dns default_dns = {
 
 static bool
 	initialized;
-
-static void __attribute__((constructor)) net_init(void)
-{
-	if (!initialized) {
-		wget_thread_mutex_init(&default_dns.mutex);
-		initialized = true;
-	}
-}
-
-static void __attribute__((destructor)) net_exit(void)
+static void net_exit(void)
 {
 	if (initialized) {
 		wget_thread_mutex_destroy(&default_dns.mutex);
 		initialized = false;
+	}
+}
+
+INITIALIZER(net_init)
+{
+	if (!initialized) {
+		wget_thread_mutex_init(&default_dns.mutex);
+		initialized = true;
+		atexit(net_exit);
 	}
 }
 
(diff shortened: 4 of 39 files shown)


libwget/hash_printf.c
```diff
@@ -64,16 +64,17 @@ void wget_hash_printf_hex(wget_digest_algorithm algorithm, char *out, size_t out
 	va_end(args);
 
 	if (plaintext) {
-		unsigned char digest[wget_hash_get_len(algorithm)];
+		size_t digestLen = wget_hash_get_len(algorithm);
+		unsigned char * digest = malloc(digestLen * sizeof(char));
```
gvanem, line 68: Why not simply `alloca()`?

examples/batch_loader.c
```diff
@@ -20,7 +20,7 @@
  * Read URLs from stdin and download into results/domain/.
  *
  */
-
+#include "config.h"
 #include <stdio.h>
```
gvanem, line 24: Option `-FI./config.h` could also be used.

libwget/http.c
```diff
@@ -29,7 +29,28 @@
  * RFC 6265
  *
  */
-
+#ifdef __cplusplus
```
gvanem, line 32: The same one more time! Move to an *internal place* IMHO.

libwget/init.c
```diff
@@ -24,14 +24,40 @@
  * 18.01.2013  Tim Ruehsen  created
  *
  */
-
+#ifdef __cplusplus
```
gvanem, line 27: And here,

libwget/hash_printf.c
mitchcapper: 100%, happy to go through all the changes and swap them to alloca if you folks are open to potentially changing away from variable length arrays

examples/batch_loader.c
mitchcapper: Will do, will add to the examples/Makefile.am

libwget/console.c
```diff
@@ -99,7 +99,23 @@ void wget_console_reset_fg_color(void)
...
+	DWORD mode = 0;
+	if (handle == INVALID_HANDLE_VALUE)
+		return mode;
+	if (!GetConsoleMode(handle, &mode))
+		return mode;
+	DWORD orig = mode;
+	if (is_input)
+		mode |= ENABLE_VIRTUAL_TERMINAL_INPUT;
+	else
+		mode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
```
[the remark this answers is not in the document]
mitchcapper: Correct, although right now we don't seem to do any checks to if VT is supported.  I assume if it didn't take effect we want to turn progress=none unless the force command is used right?

libwget/http.c
mitchcapper: I am not 100% sure what you mean here.   I would assume I should move the assist macros to a common internal header we can just include, I assume the "private.h" potentially?

libwget/hash_printf.c
rockdaboot: IMO there is no technical difference between VLA and alloca(), the latter feels just a bit clumsy. Or does alloca() check for stack boundaries ?

@mitchcapper Is there a technical reason to use MSVC ? I am not a windows user, but know there are ports of llvm/clang available, which seems to be a good choice for C projects. Not sure if there are usable ports of gcc.
The reason I am asking is that I am personally not a fan of changing code because of issues that a (proprietary) compiler has. I'd like to focus on gcc and clang mostly because I can easily test with those.

mitchcapper: MSVCs position on VLAs is "Variable length array (VLA) support isn't planned. VLAs provide attack vectors comparable to gets, which is deprecated and planned for removal"  per [https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170#:~:text=Variable%20length%20array%20(VLA)%20support%20isn%27t%20planned.%20VLAs%20provide%20attack%20vectors%20comparable%20to%20gets%2C%20which%20is%20deprecated%20and%20planned%20for%20removal.) .

There is no doubt the code is not as clean.   Microsofts recommendation would be to use [_malloca](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/malloca?view=msvc-170) .  I am happy to add such code (and can gate the code to be MSVC specific still using VLAs for any other compiler), but still will ugly things up a bit.  

> IMO there is no technical difference between VLA and alloca(), the latter feels just a bit clumsy. Or does alloca() check for stack boundaries ?

_malloca does have have the benefit of object alignment and some additional checks.  In addition it falls back onto the heap if space requires it, although it does result in me having to add a _freea at the end of each function (which noops if it wasn't heap allocated).   

> @mitchcapper Is there a technical reason to use MSVC ?

While I have spent my fair share of time with gdb/lldb, the VS debugging experience I have always found to be the most pleasant.  For some of the issues I ran into I am not sure I would have had the patience to track them using LLDB.  Granted I am sure for many it is the exact opposite and completely understand the desire to not support MSVC.  It was fairly clear the choice here was to go in that direction.  There are perfectly good versions of those compilers on windows, and WSL is something many people use heavily without issue.     Here the code wasn't that far from being MSVC friendly again so I figured I would see what the temperature was on bringing support back.   If not, I will just maintain a separate patch for the WIN64LinuxBuild as long as relatively easy to do so.

rockdaboot: And please check the return value of malloca (might be NULL) :).

examples/Makefile.am
```diff
@@ -14,7 +14,7 @@ noinst_PROGRAMS = \
 
 
 AM_CFLAGS = $(WERROR_CFLAGS) $(WARN_CFLAGS) -DMALLOC_RETURNS_NONNULL
-AM_CPPFLAGS = -I$(top_srcdir)/include/wget \
```
rockdaboot, line 17: The examples do not need gnulib (`lib/`). If there is any issue with Windows/MSVC, please use `#ifdef`.

libwget/dns.c
```diff
@@ -19,7 +19,28 @@
  *
  * resolver routines
  */
-
+#ifdef __cplusplus
```
rockdaboot, line 22: Do we really need this ? AFAIR, the constructors / destructors have been added to not need to call e.g. wget_http_init()/wget_http_exit() explicitly. That has a small race condition - the init() functions are not thread-safe. Introducing C11 atomics here would likely serve the purpose (does MSVC support C11 atomics ?).

(*) We do not call dns_init() from wget_dns_init()... this looks like an inconsistence that needs to be fixed.

libwget/console.c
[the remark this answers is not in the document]
rockdaboot: > I assume if it didn't take effect we want to turn progress=none unless the force command is used right?

Sounds good to me.

src/wget.c
```diff
@@ -3128,7 +3128,7 @@ static void set_file_mtime(int fd, int64_t modified)
 	timespecs[1].tv_sec = tt;
 	timespecs[1].tv_nsec = 0;
 
-	if (futimens(fd, timespecs) == -1)
+	if (! isatty(fd) && futimens(fd, timespecs) == -1)
```
rockdaboot, line 3139: Thanks you! Cherry-picked this commit.

libwget/ssl_wolfssl.c
```diff
@@ -69,6 +69,9 @@
 #include "private.h"
 #include "net.h"
 #include "filename.h"
+#ifdef _WIN32
```
rockdaboot, line 77: Thanks you ! Cherry-picked and slightly amended.

rockdaboot:
@mitchcapper I found *some* time to cherry-pick several of your commits (did some smaller amendments). Great work and great findings by the way!

By doing that, I am stretching the FSF rules a bit as the changes sum up to much more than the 15 lines of code that are considered "trivial" =).

Did you get a chance to send the Copyright assignment already ?

Anyway, I'll continue picking+merging commits over time, if you don't mind.

mitchcapper:
> Did you get a chance to send the Copyright assignment already ?
I did get all the copies and I will have then mailed in this weekend.

> Anyway, I'll continue picking+merging commits over time, if you don't mind.
Of course


I should have some of the other revisions done shortly.  There is still a bug in the core gnulib poll code im trying to make sure is resolved but should have some time on Sunday

libwget/io.c
```diff
@@ -433,12 +433,13 @@ int wget_update_file(const char *fname,
 		if (fp) {
 			// read fname data
 			if (load_func(context, fp)) {
-				fclose(fp);
+				if (! fclose(fp))
+					fp = NULL;
```
rockdaboot, line 453: `fp` is a local variable and we return two lines below here. No need to set `fp` to `NULL`.

libwget/io.c
```diff
@@ -433,12 +433,13 @@ int wget_update_file(const char *fname,
...
 				close(lockfd);
 				return WGET_E_UNKNOWN;
 			}
-
-			fclose(fp);
+			if (fp)
```
rockdaboot, line 457: We already know that `fp` is set.

PR #282, opened 2023-07-28
Add better default cert/bundle paths for Windows, and de-dupe paths out of ssl_* classes to utils

mitchcapper:
Broken out from #280 and rebased

### DIFF
--- a/libwget/utils.c
+++ b/libwget/utils.c
@@ -34,8 +34,10 @@
 #include <time.h>
 #include <glob.h>
 
+#include <stdbool.h>
 #include "c-ctype.h"
 #include "c-strcase.h"
+#include "filename.h"
 
 #if defined __clang__
   // silence warnings in gnulib code
@@ -62,6 +64,38 @@
  * They may be useful to other developers that is why they are exported.
  */
 
+static const char * ssl_default_path(bool bundleNotDir)
+{
+#ifndef _WIN32
+	return bundleNotDir ? NULL : "/etc/ssl/certs";
+#else
+	static char CERTDIR_PATH[MAX_PATH] = {0};
+	static char CERTBUNDLE_PATH[MAX_PATH] = { 0 };
+	char* buffer = bundleNotDir ? CERTBUNDLE_PATH : CERTDIR_PATH;
+	if (buffer[0]) //we could end up with a partial or incorrect path with multiple threads
+		return buffer;
+	wget_strscpy(buffer, "/etc/ssl/certs", MAX_PATH);
+
+	if (access(buffer, F_OK)) {
+		const char* progData = getenv("ProgramData");
+		if (!progData)
+			progData = "/ProgramData";
+		const char* dir_separator_str = (char[]){ DIR_SEPARATOR, '\0' };
+		sprintf_s(buffer, MAX_PATH, "%s%s%s%c%s%s", progData, ISSLASH(progData[strlen(progData - 1)]) ? "" : dir_separator_str, "ssl", DIR_SEPARATOR, bundleNotDir ? "ca-bundle.pem" : "certs", bundleNotDir ? "" : dir_separator_str);
+	}
+	return buffer;
+#endif
+}
+const char* wget_ssl_default_cert_dir()
+{
+	return ssl_default_path(false);
+}
+
+const char* wget_ssl_default_ca_bundle_path()
+{
+	return ssl_default_path(true);
+
+}
 /**
  * \param[in] s1 String
  * \param[in] s2 String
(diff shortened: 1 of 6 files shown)


libwget/utils.c
```diff
@@ -62,6 +63,37 @@
  * They may be useful to other developers that is why they are exported.
  */
 
+static const char * _wget_ssl_default_path(BOOL bundleNotDir)
```
rockdaboot, line 66: Use `bool` here - needs to include `<stdbool.h>`.

libwget/utils.c
```diff
@@ -62,6 +63,37 @@
...
+		if (!progData)
+			progData = "/ProgramData";
+		const char* dir_separator_str = (char[]){ DIR_SEPARATOR, '\0' };
+		sprintf_s(buffer, MAX_PATH, "%s%s%s%c%s%s", progData, ISSLASH(progData[strlen(progData - 1)]) ? "" : dir_separator_str, "ssl", DIR_SEPARATOR, bundleNotDir ? "ca-bundle.pem" : "certs", bundleNotDir ? "" : dir_separator_str);
+	}
+	return buffer;
+#endif
+}
+const char* wget_ssl_default_cert_dir() {
+	return _wget_ssl_default_path(FALSE);
```
rockdaboot, line 89: Use lowercase `true` here.

libwget/utils.c
```diff
@@ -62,6 +63,37 @@
...
+}
+
+const char* wget_ssl_default_ca_bundle_path()
+{
+	return _wget_ssl_default_path(TRUE);
```
rockdaboot, line 94: Use lowercase `true` here.

libwget/utils.c
```diff
@@ -62,6 +63,37 @@
...
+{
+#ifndef _WIN32
+	return bundleNotDir ? NULL : "/etc/ssl/certs";
+#else
+	static char CERTDIR_PATH[MAX_PATH] = {0};
+	static char CERTBUNDLE_PATH[MAX_PATH] = { 0 };
+	char* buffer = bundleNotDir ? CERTBUNDLE_PATH : CERTDIR_PATH;
+	if (buffer[0]) //we could end up with a partial or incorrect path with multiple threads
+		return buffer;
+	strcpy_s(buffer, MAX_PATH, "/etc/ssl/certs");
```
rockdaboot, line 76: Can we use `wget_strscpy()` here ?

libwget/utils.c
rockdaboot, line 66: Please rename to `ssl_default_path`, no need for name-spacing with "wget_", and a leading underscore is reserved in C.

libwget/utils.c
```diff
@@ -62,6 +63,37 @@
  * They may be useful to other developers that is why they are exported.
  */
 
+static const char * _wget_ssl_default_path(BOOL bundleNotDir)
+{
+#ifndef _WIN32
+	return bundleNotDir ? NULL : "/etc/ssl/certs";
+#else
+	static char CERTDIR_PATH[MAX_PATH] = {0};
```
rockdaboot, line 71: A library function should not use `static` variables - except they are only initialized once during initialization phase.
Can we consider `getenv("ProgramData")` to be a constant ? If not, we should better pass this as a function argument. If yes, the function can be called just once in `tls_init()`.

src/options.c
```diff
@@ -1223,6 +1223,9 @@ static int print_plugin_help(WGET_GCC_UNUSED option_t opt,
 }
 
 // default values for config options (if not 0 or NULL)
+/*
```
rockdaboot, line 1226: Please use a C++ style comment here, just for consistency.

libwget/utils.c
mitchcapper: I think assuming programdata is a a constant is just fine, especially as I don't think anyone expects to be able to externally edit the ENV vars of a running program.   I favored initializing on demand as even if called simultaneously there was no race condition but will move the call to tls_init instead.

mitchcapper:
other changes all made but for the initial path setting is there a place better than tls_init to do so? I can do it there but it means adding a call in all ssl engine files to that.

libwget/utils.c
```diff
@@ -62,6 +64,38 @@
...
 
+static const char * ssl_default_path(bool bundleNotDir)
+{
+#ifndef _WIN32
+	return bundleNotDir ? NULL : "/etc/ssl/certs";
+#else
+	static char CERTDIR_PATH[MAX_PATH] = {0};
+	static char CERTBUNDLE_PATH[MAX_PATH] = { 0 };
+	char* buffer = bundleNotDir ? CERTBUNDLE_PATH : CERTDIR_PATH;
+	if (buffer[0]) //we could end up with a partial or incorrect path with multiple threads
```
rockdaboot, line 75: This check can never be true

rockdaboot: Nevermind, should go to bed :)

rockdaboot:
I pushed your commit to main, but somewhat amended. Please check if it builds on Windows and works as expected.

PR #283, opened 2023-07-28
Enable native VT support on windows (fixes cmdhost etc usages)

mitchcapper:
disable progress bar if VT mode is not supported (unless force is used)

Broken out from #280 and rebased 
added disabling of the progress bar if no VT support is detected.


rockdaboot:
Thank you :+1:
Amended and merged manually.

PR #284, opened 2023-07-28
gcc specific C constructor/destructors to more generic with MSVC support

mitchcapper:
broken off of #280 and rebased
OK if this constructor code can be reviewed.   The only compiler it is different for is MSVC.  I did test multiple per executable/dll/library without issues.  I didn't see a way around the fact the initializer function name has to be unique per executable.  Without the linker comment it can get optimized out, but it needs a unique symbol name to work.  I played around with things like __COUNTER__ to dynamically generate it but that is only unique per object not per executable.

If we ever included more than one SSL engine for example the tls_init's would collide. 

I don't know the deep inner workings of the MSVC compiler to ensure this is exactly what we want (especially given some of this functionality is not documented).  It does seem reliable from the tests I have done.

### DIFF
--- a/libwget/ssl_openssl.c
+++ b/libwget/ssl_openssl.c
@@ -124,9 +124,16 @@ static int ssl_userdata_idx;
 /*
  * Constructor & destructor
  */
-static void __attribute__ ((constructor)) tls_init(void)
+static void tls_exit(void)
 {
-	if (!mutex)
+	if (mutex){
+	CRYPTO_free_ex_index(CRYPTO_EX_INDEX_APP, ssl_userdata_idx);
+		wget_thread_mutex_destroy(&mutex);
+	}
+}
+INITIALIZER(tls_init)
+{
+	if (!mutex){
 		wget_thread_mutex_init(&mutex);
 
 	ssl_userdata_idx = CRYPTO_get_ex_new_index(
@@ -136,14 +143,10 @@ static void __attribute__ ((constructor)) tls_init(void)
 		NULL,     /* dup_func */
 		NULL      /* free_func */
 	);
+	atexit(tls_exit);
+	}
 }
 
-static void __attribute__ ((destructor)) tls_exit(void)
-{
-	CRYPTO_free_ex_index(CRYPTO_EX_INDEX_APP, ssl_userdata_idx);
-	if (mutex)
-		wget_thread_mutex_destroy(&mutex);
-}
 
 /*
  * SSL/TLS configuration functions
--- a/libwget/dns.c
+++ b/libwget/dns.c
@@ -69,19 +69,20 @@ static wget_dns default_dns = {
 static bool
 	initialized;
 
-static void __attribute__((constructor)) dns_init(void)
-{
-	if (!initialized) {
-		wget_thread_mutex_init(&default_dns.mutex);
-		initialized = true;
-	}
-}
-
-static void __attribute__((destructor)) dns_exit(void)
+static void dns_exit(void)
 {
 	if (initialized) {
 		wget_thread_mutex_destroy(&default_dns.mutex);
 		initialized = false;
+	}
+}
+
+INITIALIZER (dns_init)
+{
+	if (!initialized) {
+		wget_thread_mutex_init(&default_dns.mutex);
+		initialized = true;
+		atexit(dns_exit);
 	}
 }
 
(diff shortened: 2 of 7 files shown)


rockdaboot:
> If we ever included more than one SSL engine for example the tls_init's would collide.

Yes, let's address this when/if this ever happens.

What problems did you exactly run into with an MSVC compiled wget2 ?

Because the code should be prepared and work well even if the `constructor/destructor` attributes are ignored by the compiler. E.g. see `wget_ssl_init()`... it explicitly calls `tls_init()` and the documentation says
```
 * On systems with automatic library constructors/destructors, this function
 * is thread-safe. On other systems it is not thread-safe.
```

So wget2 isn't affected. And if anyone else builds a tool around libwget that calls `wget_ssl_init()` in concurrent threads (bad programming pattern btw), it should use its own locking around the call.

So I wonder if there is anything to fix at all, and if so, why.

mitchcapper:
The only problems I ran into were the undefined __attribute__ constructor calls.    I simply tried to find a solution to emulate it so no matter the compiler it would behave the same.  Are you just suggesting not using constructors/destructors at all?

rockdaboot:
> Are you just suggesting not using constructors/destructors at all?

For WIN32, yes. I mean, it is an option that already should work. MSVC is possibly not the only compiler that doesn't support constructors/destructors. I thought about this when inventing the constructors/destructors but had no way to really test it. Maybe you can test it and figure out what goes wrong (if something goes wrong).

mitchcapper:
might be a stupid question, but if the code should always work without any benefit to using them based on compiler choice?  is it to potentially avoid race conditions by calling code or trading startup performance for first request perf? One option could be to disable constructor/destructor calls except for release builds (so more likely to find any bugs during dev but not introduce potential race conditions on anything using it in production.

I will say the MSVC code was not something I pulled out of thin air I looked at the slew of gnu related on my local drive and some large GH projects.  The counter based unique function construction I didn't see anywhere but I cut that from the final.

- jetBrains/kotlin(45295*) https://github.com/JetBrains/kotlin/blob/master/kotlin-native/runtime/src/mimalloc/c/init.c#L564-L580
- hasura/graphql-engine (30018*) https://github.com/hasura/graphql-engine/blob/master/preload-mimalloc/mimalloc/src/init.c#L654-L685
- ceph/ceph (12183*) https://github.com/ceph/ceph/blob/main/src/include/compat.h#L392-L402
- monero-project/monero (7997*) https://github.com/monero-project/monero/blob/master/src/crypto/initializer.h#L43-L60
- kuno/GeoIP (415*) https://github.com/kuno/GeoIP/blob/master/deps/GeoIP-1.6.7/libGeoIP/pread.c#L31-L36

Along with testing the multiple static/dynamic/library/internal combos to make sure all sets of constructors were called.

rockdaboot:
> One option could be to disable constructor/destructor calls except for release builds (so more likely to find any bugs during dev but not introduce potential race conditions on anything using it in production.

That is a dangerous option. Maybe all works fine in debug mode, but in production not. Better the functionality is the same for debug and prod.

libwget/ssl_openssl.c
```diff
@@ -124,7 +124,13 @@ static int ssl_userdata_idx;
 /*
  * Constructor & destructor
  */
-static void __attribute__ ((constructor)) tls_init(void)
+static void tls_exit(void)
+{
+	CRYPTO_free_ex_index(CRYPTO_EX_INDEX_APP, ssl_userdata_idx);
```
rockdaboot, line 129: This shouldbe protected by `if (mutex) ...`

libwget/ssl_openssl.c
```diff
@@ -136,14 +142,9 @@ static void __attribute__ ((constructor)) tls_init(void)
 		NULL,     /* dup_func */
 		NULL      /* free_func */
 	);
+	atexit(tls_exit);
```
rockdaboot, line 145: Everything here should be protected by `if (!mutex)...`

rockdaboot:
Is there C11 support in (latest) msvc? I consider using <stdatomic.h> for setting the `initialized` variables atomically to protect against race conditions and not using a mutex.
(It is unrelated to this PR.)

mitchcapper:
Fixed constructor for openssl.  Given all the factors I would recommend using this solution for MSVC rather than none only because it is a minor platform for wget2 compiling.  This would give it the best chance to behave like gcc does for similar code.

rockdaboot:
Thank you!
Slightly amended and pushed to master.

PR #285, opened 2023-07-28
Don't enable debug logging with wolfssl unless defined prior to including the options file.

mitchcapper:
If you compile WOLFSSL with debug support it flips the debug mode on by default in its options file.  This code automatically flips that off unless we pass the DEBUG_WOLFSSL define  when compiling wget2 as well.

### DIFF
--- a/libwget/ssl_wolfssl.c
+++ b/libwget/ssl_wolfssl.c
@@ -45,7 +45,12 @@
 #  define SOCKET_TO_FD(x) (x)
 #endif
 
+#ifndef DEBUG_WOLFSSL
 #include <wolfssl/options.h>
+#undef DEBUG_WOLFSSL
+#else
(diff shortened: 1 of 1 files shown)


libwget/ssl_wolfssl.c
```diff
@@ -30,6 +30,12 @@
  * https://test-sspev.verisign.com:2443/test-SSPEV-revoked-verisign.html
  *
  */
+#ifndef DEBUG_WOLFSSL
+#include <wolfssl/options.h>
```
rockdaboot, line 34: In L54, `wolfssl/options.h` is included a second time. Can you move the code there ?

mitchcapper:
all set

rockdaboot:
Thank you. Merged manually.

PR #286, opened 2023-07-28
Switch away from c99 variable arrays to malloca

mitchcapper:
re-enables msvc support

torn off #280 and rebased.  Added null return checking as well. Note:

I started off trying to mimic the failure code around the malloca call to do similar incase of memory failure but honestly error_printf_exit may just be better everywhere? I am not sure continuing after alloc failure is a use case one needs to support but open to other ideas.

### DIFF
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -82,6 +82,7 @@ listen
 lock
 maintainer-makefile
 malloc-posix
+malloca
 memchr
 mkdir
 mkstemp
--- a/libwget/bar.c
+++ b/libwget/bar.c
@@ -37,6 +37,7 @@
 #include <signal.h>
 #include <wchar.h>
 
+#include <malloca.h>
 #include <wget.h>
 #include "private.h"
 
@@ -653,10 +654,14 @@ void wget_bar_print(wget_bar *bar, int slot, const char *display)
  */
 void wget_bar_vprintf(wget_bar *bar, int slot, const char *fmt, va_list args)
 {
-	char text[bar->max_width + 1];
+	size_t textSize = (bar->max_width + 1);
+	char * text = malloca(sizeof(char) * textSize);
+	if (! text)
+		error_printf_exit(_("Allocation failure of malloca\n"));
 
-	wget_vsnprintf(text, sizeof(text), fmt, args);
+	wget_vsnprintf(text, textSize, fmt, args);
 	wget_bar_print(bar, slot, text);
+	freea(text);
 }
 
 /**
--- a/libwget/hash_printf.c
+++ b/libwget/hash_printf.c
@@ -29,6 +29,7 @@
 #include <stddef.h>
 
 #include <wget.h>
+#include <malloca.h>
 #include "private.h"
 
 /**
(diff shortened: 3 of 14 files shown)


rockdaboot:
I started experimenting with using `malloca` and ran into gcc and clang warnings about the use of `alloca` (`malloca` is just a macro using `alloca` and `malloc`). The reason is that an unbounded `alloca` is as dangerous as an unbounded VLA. And the compilers are too dump to recognize that `alloca` is bounded. I should have seen this earlier to not waste your time :|, please excuse.

I also wonder about the MS suggestion to avoid VLA and instead use `alloca`.

Anyway, looking at each of the VLAs, I think each one needs a different "proper" solution.
If you don't mind, I'll take over and address the VLAs without using `malloca` and `alloca`.
Libwget already has several "use stack and fall back to heap if needed" patterns.

That should reduce the overall developer hours as we don't have to iterate on every detail.
Does this sound OK to you ?

mitchcapper:
No worries sounds great.

mitchcapper:
Minor note as well, MS  does _not_ recommend alloca they specifically recommend malloca because of the fact it falls back onto the heap for calls that are too large.

rockdaboot:
Can you check if latest master compiles with msvc (in regards to VLAs)?

gvanem:
> Can you check if latest master compiles with msvc (in regards to VLAs)?

I have compiled using MSVC and have no issues regarding VLA. But `cl` has serious issues with `options.c`:
```c
src/options.c(263,1): error C2121: '#': invalid character: possibly the result of a macro expansion
        );
^
src/options.c(147,2): error C2143: syntax error: missing ')' before 'if'
        puts("+digest"
        ^
src/options.c(147,2): error C2059: syntax error: ')'
        puts("+digest"
        ^
```

Since the pre-processor in MSVC is so stupid, the version stuff should be printed as Wget does in 
it's generated `build_info.c`. Something like:
```c
static int print_version(WGET_GCC_UNUSED option_t opt, WGET_GCC_UNUSED const char *val, WGET_GCC_UNUSED const char invert)
{
#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
	puts("GNU Wget2 " PACKAGE_VERSION " - multithreaded metalink/file/website downloader\n");

const char *version_info[] = {
        "+digest"
#if defined WITH_GNUTLS
	" +https"
	" +ssl/gnutls"
#elif defined WITH_OPENSSL
	" +https"
	" +ssl/openssl"
#elif defined WITH_WOLFSSL
	" +https"
	" +ssl/wolfssl"
#else
	" -https"
	" -ssl"
#endif
    ...
  };
  puts(version_info);
  ...
```

rockdaboot:
Could you please test whether branch `tim/cl-version-info` compiles with msvc ?

mitchcapper:
Sorry can you clarify where that branch is? I don't see cl-version-info under this repo.  Given you are tim I checked @rockdaboot/wget2/cl-version-info   also not.   Is it on gitlab or am I missing it?

rockdaboot:
> Is it on gitlab or am I missing it?

Ah sorry, it is on Gitlab. The Github repo is just a mirror (but nevermind).

rockdaboot:
The most important thing is that the CI is running on Gitlab. So if you open MRs/PRs on Gitlab, they are automatically checked (the MinGW64 runner is broken currently).

PR #287, opened 2023-07-28
Add config.h/libgnu to all example compiles for gnulib

mitchcapper:
in #280 you mentioned these additions be gated to _WIN32  but I haven't quite done so.   Mainly because if you include a gnulib header accidentally you can end up getting the config.h missing error and/or need the libgnu library itself to avoid symbol issues.   I assume the issue isn't the minor overhead as much as having a clearer sample for others?  With this the code on the primary case base and example code base should behave the same without one might notice differences.

If you want to confirm the preference for these additions to be WIN32 only I will get the conditionals added to the makefile.


rockdaboot:
The examples should be absolutely independent of gnulib. They should only depend on libwget and libc (or maybe some other external libraries).

They serve as an example on how to use libwget in other projects. `gnulib` is internal to libwget, wget2 and the tests.

mitchcapper:
no problem will update.

mitchcapper:
Done.   Conditioned on MSVC as that is when one would not have unistd.h without gnulib.

rockdaboot:
IMO, this is the wrong approach. Now, building the examples do not only require libwget but also libgnu. And the latter is not available to tool developers outside the wget2 project.

On MinGW64 (and on Linux), the needed gnulib code is included in libwget, so no extra library is needed. Also, `config.h` is not needed.

To me it looks like the library builds are broken when using MSVC. Do you have any idea how to solve this?

mitchcapper:
No problem, let me look at the examples.  It immediately failed as https://github.com/rockdaboot/wget2/blob/14d0460597ab98a519af5396d91bb430256023e6/examples/print_css_urls.c#L31   and unistd doesn't exist in windows.     My immediate approach to any compat is throw gnulib at it, but your point is right the examples we provide shouldn't require it.   If I can just get the headers out of the way that are problematic I will do that, if the functions required are minor ill see about just adding windows shims for it, or if worse case  can always disable the ones that wouldn't work on win32 without gnulib if another solution doesn't.

rockdaboot:
:+1: Sounds good to me.

rockdaboot:
@mitchcapper Can we close this?

I don't remember if we found a solution, e.g. conditionally skipping includes on WIN32/MSVC. But if there is still something to do, please open another PR.

Issue #311, opened 2024-04-17
bug(windows): stdin/stdout modes of the calling process is not restored after wget2 exits


insomnimus:
On Windows, wget2 currently sets the console modes via `SetupConsoleHandle` in libwget (`console.c`) so that stdin gets `ENABLE_VIRTUAL_TERMINAL_INPUT`, and stdout `ENABLE_VIRTUAL_TERMINAL_PROCESSING`.

This seems to cause problems after wget2 exits (noticeably with input) on the default terminal conhost.exe, as well as the new Windows Terminal and likely others:
- Pressing `ctrl+enter`, instead of doing whatever action it should, inputs the character `j` into the terminal
- Pressing `ctrl+backspace`, instead of deleting to the left of the word, deletes a single character
- Pressing `ctrl+f` inputs the character `F`
- Customly assigned keybinds such as `ctrl+1` do nothing, including their action

These are just a couple - the problem isn't specific to any of these key combinations.

The solution I've figured out to this problem is restoring the initial console mode as obtained with `GetConsoleMode` before the program exits.
I've tested that restoring the initial modes before exit indeed solves the problem.

Since the console mode's set in libwget, ideally it should also be restored by libwget (maybe something to be called by the library user before exit).

rockdaboot:
Fixed in master

PR #312, opened 2024-04-17
fix(windows): restore console modes before exit

insomnimus:
On Windows, the console modes are altered by wget2 (via `SetupConsoleHandle`in libwget), to enable virtual terminal sequences.
However it's not restored before exit; this causes problems such as key binds not working properly, even after the wget2 process exits (#311).

This PR addresses the problem by saving the console modes at the start of main, and then restoring them before main returns.

I'm not well versed in C let alone this code base, so the PR might not be up to the project's standards. Do let me know if I should change anything.

Fixes: #311

### DIFF
--- a/libwget/init.c
+++ b/libwget/init.c
@@ -50,10 +50,32 @@ static int global_initialized;
 static wget_thread_mutex _mutex;
 static bool initialized;
 
+#ifdef _WIN32
+	// The initial stdin and stdout modes, to be restored before exit.
+	static DWORD initial_stdin_mode = 0;
+	static DWORD initial_stdout_mode = 0;
+
+	BOOL HandleCtrlEvent(DWORD dwCtrlType)
+	{
+		// We exit with 2 per convention, as this is similar to SIGINT on UNIX.
+		// The destructor registered with atexit will do the cleanup work.
+		// If it's not ctrl+c nor ctrl+break, we return FALSE so other handlers, if any, get to run.
+		if(dwCtrlType == CTRL_C_EVENT || dwCtrlType == CTRL_BREAK_EVENT)
+			exit(2);
+		else
+			return FALSE;
+	}
+#endif
+
 static void global_exit(void)
 {
 	if (initialized) {
 		wget_thread_mutex_destroy(&_mutex);
+		#ifdef _WIN32
+			// Restore console modes.
+			SetConsoleMode(GetStdHandle(STD_INPUT_HANDLE), initial_stdin_mode);
+			SetConsoleMode(GetStdHandle(STD_OUTPUT_HANDLE), initial_stdout_mode);
+		#endif
 		initialized = false;
 	}
 }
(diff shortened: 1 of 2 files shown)


gvanem:
What if a `CTRL_C_EVENT` or `CTRL_BREAK_EVENT` gets raised?
AFAICS, there's no call to `SetConsoleCtrlHandler()` to trap and restore the terminal-state.

gvanem:
>  It still ends up restoring console modes.

Are you sure? Neither `SetConsoleCtrlHandler()` nor `SetConsoleMode()` are called in Wget2.
And there's a difference in how `conhost.exe` handles Ctrl-C vs. Ctrl-Break:
https://learn.microsoft.com/en-us/windows/console/ctrl-c-and-ctrl-break-signals

insomnimus:
Ah, you're right: `ctrl-break` still leaves the console modified; however `ctrl-c` does not.

So the solution is `SetConsoleCtrlHandler`, however since it takes a handler function, how should I go about not duplicating the cleanup code at the end of main? Also it looks like the previous console modes have to become globals now, which I'm not a fan of.

As a side note I'm still not sure why the patch works for `ctrl-c`.

insomnimus:
Also `SetConsoleMode` is actually called indirectly:
- `init` in `src/options.c`
- calls `log_init` in `src/log.c`
- calls `wget_console_init` in `libwget/console.c`
- calls `SetupConsoleHandle` also in `libwget/console.c`
- which finally calls `SetConsoleMode`

That's how i diagnosed the issue at first: commenting out `SetConsoleMode` calls inside `SetupConsoleHandle`, which worked but is of course undesirable.

insomnimus:
Are there any suggestions on how to proceed?

rockdaboot:
I think you should add your code to the global [constructor](https://github.com/rockdaboot/wget2/blob/761ae5082dae8a0b58727cd9e10dffd32cf8d03b/libwget/init.c#L60) and the global [destructor](https://github.com/rockdaboot/wget2/blob/761ae5082dae8a0b58727cd9e10dffd32cf8d03b/libwget/init.c#L53).

The destructor should be called at program exit. Please test it and give a feedback.

insomnimus:
Sorry for the delay; I got hit with the seasonal flu.

I've moved the cleanup to `libwget/init.c` as you suggested:
- The initial console modes are obtained and stored in `INITIALIZER()`.
- A handler is registered via `SetConsoleCtrlHandler()`, also in `INITIALIZER()`.
- The handler simply calls `exit(2)` if the signal is `ctrl+c` or `ctrl+break`; it ignores other signals, letting any other handler, if any, handle them. This triggers whatever's registered with `atexit()`, including our destructor `global_exit()`.
- The console modes are restored in `global_exit()`.

Let me know if I should change anything.

insomnimus:
Oh also I've tested this and it works for both `ctrl+c` and `ctrl+break`.

rockdaboot:
Thanks for your contribution, manually pushed (changed the commit message).

Issue #342, opened 2024-09-11
Error checking for SSL_read and SSL_write is off by one which can lead to an infinite loop


tobiasb-ms:
NOTE: I only encounter this when I see an odd networking issue on a VM that I'm still root causing. However, at first pass it does seem to be a bug in `wget2` that the networking issue is surfacing.

In `ssl_openssl.c`, in the function [ssl_transfer](https://github.com/rockdaboot/wget2/blob/ec27488feadd44b5e126592bc18ff2441f8cae5a/libwget/ssl_openssl.c#L1786), the check on the return value from `SSL_read` or `SSL_write` (depending on `want` parameter) [considers 0 a success](https://github.com/rockdaboot/wget2/blob/ec27488feadd44b5e126592bc18ff2441f8cae5a/libwget/ssl_openssl.c#L1819). The [docs ](https://docs.openssl.org/3.3/man3/SSL_read/#return-values)for [both](https://docs.openssl.org/3.3/man3/SSL_write/#return-values) say that 0 is an error.

This means that we end up returning 0 rather than going into either the retry logic or (as is appropriate for the case I've seen), [returning WGET_E_HANDSHAKE](https://github.com/rockdaboot/wget2/blob/ec27488feadd44b5e126592bc18ff2441f8cae5a/libwget/ssl_openssl.c#L1830). The calling function [wget_ssl_read_timeout](https://github.com/rockdaboot/wget2/blob/ec27488feadd44b5e126592bc18ff2441f8cae5a/libwget/ssl_openssl.c#L1859) then looks for `WGET_E_HANDSHAKE` to determine if it's an error and further up the stack, in `http2.c`, function [wget_http2_get_response_cb](https://github.com/rockdaboot/wget2/blob/ec27488feadd44b5e126592bc18ff2441f8cae5a/libwget/http2.c#L373), looks for `-1` (the value of `WGET_E_HANDSHAKE`) to break out of the loop it's in, which ends up being an infinite loop in the case I'm seeing.

I've created a draft pull request #343 as a potential fix for this.

rockdaboot:
Thanks for the report :+1:

PR #343, opened 2024-09-11
count 0 as an error for SSL_read and SSL_write, per documentation

tobiasb-ms:
Potential fix for #342.

This changes the check on the return values of `SSL_read` and `SSL_write` to consider 0 an error, per documentation.


rockdaboot:
Thanks for the contribution. I pushed a different version manually (but with same semantics).

rockdaboot:
For future PRs, please create a [Gitlab MR](https://gitlab.com/gnuwget/wget2/-/merge_requests). There we have a CI, so ideally, it involves less testing from my side :)

Issue #348, opened 2024-10-03
SIGINT shutdown getting stuck with --wait


michaelweiser:
When running wget2 in recursive download mode with `--wait` it gets stuck for me on shutdown with Ctrl+C like so:
``` sh
# wget2 --progress=none -rkl1 --wait=4 --debug http://www.gnu.org
[...]
03.164434.256 trying 2001:470:142:5::116:80...
03.164434.256 established connection www.gnu.org
^C03.164438.256 closing connection
```
I have to interrupt it a second time and it will not rewrite downloaded files to relative links (-k).

Looking at it with gdb
``` c
(gdb) info threads
  Id   Target Id                                  Frame 
* 1    Thread 0x7f9de8779ec0 (LWP 792183) "wget2" 0x00007f9de8ccacbe in ?? () from /usr/lib64/libc.so.6
(gdb) bt
#0  0x00007f9de8ccacbe in ?? () from /usr/lib64/libc.so.6
#1  0x00007f9de8ccd6a9 in pthread_cond_wait () from /usr/lib64/libc.so.6
#2  0x00007f9de92e6882 in wget_thread_cond_wait (cond=<optimized out>, mutex=<optimized out>, ms=<optimized out>) at thread.c:188
#3  0x0000561cda5ae3e7 in main (argc=<optimized out>, argv=0x7fff5336dad8) at wget.c:1494
(gdb) 
```
shows all downloader threads exited and the main thread pending in `pthread_cond_wait` here: https://github.com/rockdaboot/wget2/blob/8877050c3f00a19d43e539029d2346d1040d8c02/src/wget.c#L1494

Since it does not happen without `--wait`, my best guess is that `downloader_thread` breaks its main loop after returning from its wait sleep and finding `terminate` set here: https://github.com/rockdaboot/wget2/blob/8877050c3f00a19d43e539029d2346d1040d8c02/src/wget.c#L2464
While all other code paths seem to eventually signal the main thread for one reason or another, this one does not. So `main` goes on waiting for a signal that will never come.

Consequently, waking the main thread on exit like so cures the issue for me:
``` diff c
diff --git a/src/wget.c b/src/wget.c
index 932ce552..13cff8f7 100644
--- a/src/wget.c
+++ b/src/wget.c
@@ -2566,6 +2566,9 @@ out:
        // if we terminate, tell the other downloaders
        wget_thread_cond_signal(worker_cond);
 
+       // ... and main thread so it does not get stuck on shutdown
[...]

PR #349, opened 2024-10-03
Wake the main thread on downloader shutdown

michaelweiser:
Without this, the main thread may get stuck waiting for signals from downloaders while those have already exited and will never signal it. Observed with --wait.

Fixes: #348


rockdaboot:
Thanks for the fix/commit, LGTM.

(I have to manually apply the patch and can't just merge because this repository is a mirror).

rockdaboot:
TY. Manually merged with a slightly amended commit message.

PR #350, opened 2024-10-16
Added bzip2 support

tear-cmd:
Hello there! So i just made added support for bzip2! I tried it myself and it has been tested mutliple times. Its working flawlessly. This might not look like it is much, but i spend an hour figuring out where and how to add the support for bzip2. At the end of the day it is working.

If something that isnt suitable for http_get.c that i changed let me know!


rockdaboot:
> At the end of the day it is working.

I am sorry. At the end of the day this is not working :(

Wget2 already has bzip2 support since a long time. Just enable it at build time:
```
$ ./configure --help|grep bzip
  --with-bzip2            enable bzip2 compression support
```

Issue #356, opened 2024-11-18
-O always downloads the full file


mruprich:
When using -c and -O, the target file is always downloaded in full size. The content-range starts with 0B for every file. Reproducer:
```
# wget -c -O kernel-modules.rpm https://mirror.karneval.cz/pub/linux/fedora/linux/development/41/Everything/x86_64/os/Packages/k/kernel-debug-modules-6.11.4-301.fc41.x86_64.rpm
kernel-modules.rpm   100% [====================================================================>]  106.55M   32.00MB/s
                          [Files: 1  Bytes: 106.55M [23.06MB/s] Redirects: 0  Todo: 0  Errors: 0]
# ls -lh kernel-modules.rpm
-rw-r--r--. 1 root root 107M Oct 20 13:01 kernel-modules.rpm
# truncate --size 50M kernel-modules.rpm
# ls -lh kernel-modules.rpm
-rw-r--r--. 1 root root 50M Nov 18 07:58 kernel-modules.rpm
# wget --debug -c -O kernel-modules.rpm https://mirror.karneval.cz/pub/linux/fedora/linux/development/41/Everything/x86_64/os/Packages/k/kernel-debug-modules-6.11.4-301.fc41.x86_64.rpm
....
18.075855.798 content-range: bytes 0-111736168/111736169   <---- Always range from 0
18.075855.798 [FRAME 1] < HEADERS
kernel-modules.rpm    96% [=================================================================>   ]  103.19M   26.40MB/s
18.075855.798 [FRAME 0] > SETTINGS
kernel-modules.rpm   100% [====================================================================>]  106.55M   27.43MB/s
18.075900.377   ##  response status 206
18.075900.377 keep_alive=1
18.075900.377 _host_remove_job: 0x559e9ef57cf0
18.075900.377 host_remove_job: qsize=0 host->qsize=0
18.075900.377 [0] action=1 pending=0 host=0x559e9ef57690
18.075900.377 closing connection
18.075900.377 [0] action=1 pending=0 host=0x0
18.075900.377 main: wake up
18.075900.378 main: done
18.075900.378 queue_size: [Files: 0  Bytes: 106.55M [20.85MB/s] Redirects: 0  Todo: 0  Errors: 1]
18.075900.378 blacklist https://mirror.karneval.cz/pub/linux/fedora/linux/development/41/Everything/x86_64/os/Packages/k/kernel-debug-modules-6.11.4-301.fc41.x86_64.rpm
```
[...]

PR #357, opened 2024-11-18
src/options.c: do not truncate when -c is used

mruprich:
This should fix https://github.com/rockdaboot/wget2/issues/356.


rockdaboot:
Thank you for the contribution!
Pushed the commit manually because Gitlab is the source of truth (otherwise we may run into synchronization conflicts).

PR #361, opened 2024-11-26
Signed char for conditions

### LINKED ISSUE #360: Build fails on test-restrict-ascii on non-x86 architectures
I was preparing to build the new version of wget2 on Fedora and on aarch64, s390x and ppc64le one particular test fails - test-restrict-ascii. It most likely fails on the fact that in function wget_restrict_file_name, when parsing fname, char c has different value on x86 and different on the rest of the above mentioned architectures. In the test on x86 the file with Greek alphabet is saved like this `Saving '%CE%B3%CE%BD%E1%BF%B6%CF%83%CE%B9%CF%82.html'` but on the other arches, the file is saved as this `Saving 'γνῶσις.html'`.

The function has been edited in this commit - https://github.com/rockdaboot/wget2/commit/284954553613f75e57ef107ceaa06ae4d9dd8c59.

It has to do with the fact that char on x86 will always be unsigned but it will be signed on the other arches (at least I think). And so here, on x86, the first part of if is used but on the other arches, the else is used:
```
if ((mode_unix && (c >= 1 && c <= 31))
      || (windows && (c < 32 || strchr("\\<>:\"|?*", c)))
      || (!nocontrol && ((c >= 1 && c <= 31) || c == 127))
      || (ascii && c <= 31))
    {
      char nibble;
      *dst++ = '%';
      *dst++ = (nibble = ((unsigned char) c >> 4)) >= 10 ? nibble + 'A' - 10 : nibble + '0';
      *dst++ = (nibble = (c & 0xf)) >= 10 ? nibble + 'A' - 10 : nibble + '0';
      escaped = true;
    } else {
      *dst++ = c;
    }
```
When trying to run this in gdb, the value of c on aarch64 for the first iteration in the above condition is this:
```
156			if (lowercase && c >= 'A' && c <= 'Z') { // isupper() also returns true for chars > 0x7f, the test is not EBCDIC compatible ;-)
(gdb) p c
$10 = 206 '\316'
(gdb) p (signed char) c
$11 = -50 '\316'
```
Same lines for x86_64:
```
156			if (lowercase && c >= 'A' && c <= 'Z') { // isupper() also returns true for chars > 0x7f, the test is not EBCDIC compatible ;-)
(gdb) p c
$12 = -50 '\316'
(gdb) p (signed char)c
$2 = -50 '\316'
```

I am not yet sure how to fix this, maybe you'll be able to see the fix as something obvious right away.

mruprich:
For the purposes of the comparison of the value in c in various conditions, c needs to be signed char.

Fixes #360

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -151,7 +151,7 @@ char *wget_restrict_file_name(const char *fname, char *esc, int mode)
 	bool ascii = mode & WGET_RESTRICT_NAMES_ASCII;
 
 	for (dst = esc, s = fname; *s; s++) {
-		char c = *s;
+		signed char c = *s;
 
 		if (lowercase && c >= 'A' && c <= 'Z') { // isupper() also returns true for chars > 0x7f, the test is not EBCDIC compatible ;-)
 			c |= 0x20;


rockdaboot:
Thanks! Manually merged.

PR #363, opened 2024-11-26
Preventing truncation with --no-clobber

mruprich:
I think that in a normal situation, config.clobber is set to true so this would result in a truncated file, which is correct. In a situation when -nc is used, the file should be left as it is.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -3492,7 +3492,7 @@ int init(int argc, const char **argv)
 	if (config.output_document && strcmp(config.output_document, "-") && !config.dont_write) {
 		if (config.unlink) {
 			unlink(config.output_document);
-		} else if (!config.continue_download) {
+		} else if (!config.continue_download && config.clobber) {
 			int fd = open(config.output_document, O_WRONLY | O_TRUNC | O_BINARY);
 
 			if (fd != -1)


rockdaboot:
Hm, I have to check that wget2 behaves like wget in this case (no time right now), or did you do that already?

The docs:
```
       --no-clobber
           If  a  file  is  downloaded more than once in the same directory, Wget's behavior depends on a few options, including
           -nc.  In certain cases, the local file will be clobbered, or overwritten, upon repeated download.  In other cases  it
           will be preserved.

           When  running  Wget  without  -N,  -nc, -r, or -p, downloading the same file in the same directory will result in the
           original copy of file being preserved and the second copy being named file.1.  If that file is downloaded yet  again,
           the  third  copy  will  be  named  file.2,  and  so on.  (This is also the behavior with -nd, even if -r or -p are in
           effect.)  When -nc is specified, this behavior is suppressed, and Wget will refuse to download newer copies of  file.
           Therefore,  ""no-clobber"" is actually a misnomer in this mode---it's not clobbering that's prevented (as the numeric
           suffixes were already preventing clobbering), but rather the multiple version saving that's prevented.

           When running Wget with -r or -p, but without -N, -nd, or -nc, re-downloading a file  will  result  in  the  new  copy
           simply  overwriting  the  old.   Adding  -nc  will  prevent this behavior, instead causing the original version to be
           preserved and any newer copies on the server to be ignored.

           When running Wget with -N, with or without -r or -p, the decision as to whether or not to download a newer copy of  a
           file  depends  on  the local and remote timestamp and size of the file.  -nc may not be specified at the same time as
           -N.

           A combination with -O/--output-document is only accepted if the given output file does not exist.

           Note that when -nc is specified, files with the suffixes .html or .htm will be loaded from the local disk and  parsed
           as if they had been retrieved from the Web.

```

mruprich:
Testing quickly with a different file in wget1:

PR #364, opened 2024-11-27
Use the local system timestamp when --no-use-server-timestamps is set

mruprich:
I stumbled upon this by accident - when the --no-use-server-timestamps option is used on the command line, it is never used in the code later. I hope I used it in the right place.

Regards,
Michal Ruprich


rockdaboot:
Thank you, merged manually.

PR #366, opened 2024-12-10
wolfcrypt: Remove sslv2 support, at best was a stub and condition other older methods on WC config define

mitchcapper:
On modern wolfcrypt builds SSLv2_client_method only exists as a stub and only if #NO_OLD_TLS is not defined.    I don't think SSLV2 was actually supported any time recently by it, so I have removed it completely.  Otherwise you get a compiler error.

I also conditioned the other older ones (SSLv3->TLS V1.1) on the wolfcrypt NO_OLD_TLS flag not being defined as with it defined they are disabled per: https://www.wolfssl.com/documentation/manuals/wolfssl/chapter02.html.  This way someone doesn't try to use SSLv3 for example and not understand why its not working, now it would return missing method error.

Finally, Added an SSL alias for SSLv3 as the comment at the top mentions this is how to specify V3, (guessing openssl copy) so to make the behavior match the comment (and I assume openssl) added the alias.

Happily gets my autobuild working https://github.com/mitchcapper/WIN64LinuxBuild/actions/workflows/tool_wget2_build.yml

### DIFF
--- a/libwget/ssl_wolfssl.c
+++ b/libwget/ssl_wolfssl.c
@@ -594,10 +594,9 @@ void wget_ssl_init(void)
 
 		debug_printf("WolfSSL init\n");
 		wolfSSL_Init();
+#ifndef  NO_OLD_TLS
 
-		if (!wget_strcasecmp_ascii(config.secure_protocol, "SSLv2")) {
-			method = SSLv2_client_method();
-		} else if (!wget_strcasecmp_ascii(config.secure_protocol, "SSLv3")) {
+		if (!wget_strcasecmp_ascii(config.secure_protocol, "SSLv3") || !wget_strcasecmp_ascii(config.secure_protocol, "SSL")) {
 			method = wolfSSLv23_client_method();
 			min_version = WOLFSSL_SSLV3;
 		} else if (!wget_strcasecmp_ascii(config.secure_protocol, "TLSv1")) {
@@ -606,7 +605,9 @@ void wget_ssl_init(void)
 		} else if (!wget_strcasecmp_ascii(config.secure_protocol, "TLSv1_1")) {
 			method = wolfSSLv23_client_method();
 			min_version = WOLFSSL_TLSV1_1;
-		} else if (!wget_strcasecmp_ascii(config.secure_protocol, "TLSv1_2")) {
+		} else 
+#endif // ! NO_OLD_TLS
+			if (!wget_strcasecmp_ascii(config.secure_protocol, "TLSv1_2")) {
 			method = wolfSSLv23_client_method();
 			min_version = WOLFSSL_TLSV1_2;
 		} else if (!wget_strcasecmp_ascii(config.secure_protocol, "TLSv1_3")) {


libwget/ssl_wolfssl.c
```diff
@@ -594,10 +594,9 @@ void wget_ssl_init(void)
 
 		debug_printf("WolfSSL init\n");
 		wolfSSL_Init();
+#ifndef  NO_OLD_TLS
 
-		if (!wget_strcasecmp_ascii(config.secure_protocol, "SSLv2")) {
-			method = SSLv2_client_method();
-		} else if (!wget_strcasecmp_ascii(config.secure_protocol, "SSLv3")) {
+		if (!wget_strcasecmp_ascii(config.secure_protocol, "SSLv3") || !wget_strcasecmp_ascii(config.secure_protocol, "SSL")) {
```
rockdaboot, line 599:
```suggestion
		if (!wget_strcasencmp_ascii(config.secure_protocol, "SSL", 3)) {
```

libwget/ssl_wolfssl.c
```diff
@@ -594,10 +594,9 @@ void wget_ssl_init(void)
 
 		debug_printf("WolfSSL init\n");
 		wolfSSL_Init();
+#ifndef  NO_OLD_TLS
```
rockdaboot, line 597: The `NO_OLD_TLS` seems to be something from your build system!?
Can we just use TLSV1_2 or higher as default for "auto" and still allow users to explicitly specify older versions? Because there are cases where ancient embedded systems only understand older protocol versions. It would be tedious if you need to build your own version of wget2 just to connect to those systems.

mitchcapper: You would think but no that isn't my declare:) Wolfcrypt is a bit odd and and a massive slew of their own declares around:


NO_OLD_TLS is theirs.   If it isn't defined you won't be getting TLSV1_1 even if you want it:).

libwget/ssl_wolfssl.c
mitchcapper: I can make this change, only issue is if someone specifies SSLv2 it won't tell them invalid option but they also won't get what they ask for.  I assumed we would prefer fall to the default case of `Missing TLS method`.  Confirm this is desired instead?

libwget/ssl_wolfssl.c
rockdaboot: Thanks for the link. I wish there was a way to figure out the built-in capabilities at runtime. But here we go 🤷🏻

libwget/ssl_wolfssl.c
rockdaboot: You raise a good point. I just compared the WolfSSL code with our OpenSSL and GnuTLS implementations, that why I suggested the above. IMO, all three should behave semantically the same. But I now realize that the docs need to be changed and maybe in a follow-up PR, we need a bit of code deduplication.

So yes, please change the code as suggested above.

I take care of the docs later, and print a warning that SSLv3 is used instead of `SSLv2` (if explicitly requested).

rockdaboot:
TY!
Added same/similar functionality in master.

